PR #2, opened 2022-01-25
feat: Sign up form P2022-363

wolnio:


pages/user/register.js
```diff
@@ -1,7 +1,127 @@
...
+      errors.username = "Username must be 20 characters or less!";
+    } else if (/\W/.test(values.username)) {
+      errors.username = "Only characters and digits allowed!";
+    }
+
+    /* Email validation
+     */
+    if (!values.email) {
+      errors.email = "Email is required";
+    } else if (!/^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/.test(values.email)) {
```
MS151994, line 58: I would add pattern to the const variable.

pages/user/register.js
```diff
@@ -1,7 +1,127 @@
...
+    const errors = {};
+
+    /* Username validation
+     */
+    if (!values.username) {
+      errors.username = "Username is required";
+    } else if (values.username.length < 6) {
+      errors.username = "Username must be at least 6 characters!";
+    } else if (values.username.length > 20) {
+      errors.username = "Username must be 20 characters or less!";
```
rast-mk1, line 49: if `Username must be 20 characters or less!` then shouldn't `values.username.length > 19` ?

wolnio: Well I don't think so. If I do _x>20_ and x will be 20 then this statement will be false which means user won't get an error because it's allowed to have password with 20 characters.

rast-mk1: Yeah, my bad

pages/user/register.js
```diff
@@ -1,7 +1,127 @@
...
+
+  const handleSubmin = (e) => {
+    e.preventDefault();
+
+    setFormErrors(validate(formValues));
+    console.log(formErrors);
+    if(Object.keys(formErrors).length === 0){
+      console.log(formValues);
+      setSubmitted(true);
+      console.log(submitted);
```
marcelina-hasiak, line 35: Have you noticed that line:34 won't write you down errors messages after first form submitting? And after that is always one state 'snapshot' behind. useState hook is asynchronous and the changes won't be reflected immediately. The code needs to be improved to correctly return user credentials.

wolnio: Yeah, I noticed strange behavior but didn't think where the problem really is, thanks :)

pages/user/register.js
```diff
@@ -1,7 +1,130 @@
...
+    } else if (/\W/.test(values.username)) {
+      errors.username = "Only characters and digits allowed!";
+    }
+
+    /* Email validation
+     */
+    if (!values.email) {
+      errors.email = "Email is required";
+    } else if (
+      !/^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/.test(values.email)
```
weronika299, line 60: Isn't it better to store this RegExp in some variable?

wolnio: Yeah, I'm pretty sure it is. Someone mentioned that already, but I didn't change it yet.

PR #3, opened 2022-01-25
feat: (P2022-360) Sign up

weronika299:


pages/user/register.js
```diff
@@ -1,7 +1,99 @@
...
+    e.preventDefault();
+    setRegisterUsernameError('');
+    setRegisterPasswordError('');
+    setEmailError('');
+    setConfirmEmailError('');
+    let isErrorOccured = false;
+
+    if (!registerUsername || registerUsername.length < 5) {
+      isErrorOccured = true;
+        setRegisterUsernameError('Invalid username');
```
lenasokolova, line 36: I believe the errors should be different in this case (e.g. username is required / min 5 characters), maybe it is worth considering an if-else statement?

pages/user/register.js
```diff
@@ -1,7 +1,99 @@
...
+import { FormFlex, Input, Button } from 'components/Form';
+import { useState } from 'react';
+import styled from 'styled-components';
+
+
+const Wrapper = styled.div`
+    width: 50%;
+    margin: 50px auto;
+    background-color: ${(props) => props.theme === 'light' ? '#fff' : '#666'};
+`;
```
lenasokolova, line 10: it's better to store components inside a separate file and then just import them in order to use them in a specific file.

pages/user/register.js
```diff
@@ -1,7 +1,99 @@
...
+        error={emailError}
+        onInputChange={(e) => setEmail(e.target.value)} />
+      <Input
+        name="confirmEmail"
+        type="emaii"
+        label="Confirm Email"
+        value={confirmEmail}
+        error={confirmEmailError}
+        onInputChange={(e) => setConfirmEmail(e.target.value)} />
+      <Button onClick={userRegister}>Sign Up</Button>
```
hiimsagan, line 95: label will work without id parameter?

pages/user/register.js
```diff
@@ -1,7 +1,99 @@
...
+        setRegisterUsernameError('Invalid username');
+    } if (!registerPassword || registerPassword.length < 6) {
+      isErrorOccured = true;
+      setRegisterPasswordError('Password should be at least 6 characters long');
+    } if (confirmEmail != email) {
+      isErrorOccured = true;
+      setConfirmEmailError('Confirm email should be the same as email');
+    } if (!usernameRegEx.test(registerUsername) || registerUsername.length<6 || registerUsername.length >20) {
+      isErrorOccured = true;
+      setRegisterUsernameError('Username must be between 6 and 20 characters long and can only contain letters and numbers');
```
rast-mk1, line 45: Shouldn't it look like `if ((usernameRegEx.test(registerUsername)) && (registerUsername.length<6 || registerUsername.length >20))` instead?

pages/user/register.js
weronika299: Yes, I think you're right. Thanks! :)

pages/user/register.js
marcelina-hasiak, line 45: Now you have if statement for username in two places: please check out 34-36 lines. Maybe it's worth to merge it?

pages/user/register.js
```diff
@@ -1,7 +1,99 @@
...
+      <Input
+        name="registerPassword"
+        type="password"
+        label="Password"
+        value={registerPassword}
+        error={registerPasswordError}
+        onInputChange={(e) => setRegisterPassword(e.target.value)} />
+      <Input
+        name="email"
+        type="emaii"
```
marcelina-hasiak, line 83: Line 83 and 90: typo

pages/user/register.js
weronika299: But if someone fails the test related to regExp but the username length is valid the error should also show up. What do you think about it? 
For instance: Username has 10 characters but it has wrong  characters (regExp doesn't match) the error will not appear.

weronika299: Yes thank you. I forgot to delete those lines.

pages/user/register.js
weronika299: I need to add it. Thanks!

pages/user/register.js
weronika299: I changed it what do you think about this solution?

PR #5, opened 2022-01-25
feat: sign up #P2022-362

marcelina-hasiak:


app/components/Form/form-styling-components.js
```diff
@@ -0,0 +1,74 @@
+import styled from 'styled-components';
+
+const FormWrapper = styled.article`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  min-height: 100vh;
+  padding: 1rem;
+`;
```
wlechowicz, line 9: Used semantic article, nice!

app/components/Form/index.js
```diff
@@ -1,22 +1,24 @@
...
-            <input
-                id={props.id}
-                name={props.name}
-                type={props.type}
-                value={props.value}
-                onChange={props.onInputChange} />
-            {props.error && <div style={{ color: "red"}}>{props.error}</div>}
+            <FormRow>
+                <FormLabel isInputFocused={isInputFocused} htmlFor={id}>{label}</FormLabel>
+                <FormInput
```
wlechowicz, line 11: You changed the common component input, risky, does login still work?

wlechowicz (changes requested):
I'm worried that changing the shared `input` might have broken login. Pls check.

pages/user/login.js
```diff
@@ -8,6 +8,12 @@ const Wrapper = styled.div`
     background-color: ${(props) => props.theme === 'light' ? '#fff' : '#666'};
 `;
 
+const FormFlex = styled.form`
+    display: flex;
+    flex-direction: column; 
+    max-width: 400px;
+`;
```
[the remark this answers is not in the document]
marcelina-hasiak: Yeah! I thought about it, but then I went soft because of so many changes at `login.js` and `index.js`. But if you noticed it, I'll definitely fix it ;) Thanks!

app/components/Form/index.js
marcelina-hasiak: I've adapted `login.js` to `Input` component changes and double checked. Everything's under control ;)

pages/user/sign-up.js
```diff
@@ -0,0 +1,94 @@
...
+                username: '',
+                password: '',
+                email: '',
+                emailConfirmation: ''
+            })
+            return {
+                username,
+                password,
+                email
+            }
```
weronika299, line 48: I'm not sure if this is necessary. You set these values ​​to empty string also at the beginning.

app/utils/validateFormInputs.js
```diff
@@ -0,0 +1,33 @@
+const usernamePattern = /^(?=.*[a-zA-Z].*[a-zA-Z].*[a-zA-Z].*[a-zA-Z])[A-Za-z\d\-\_\[\]\\\/]{6,20}$/
+const emailPattern = /^[a-z\d]+[\w\d.-]*(\+[\w\d.-]*)?@(?:[a-z\d]+[a-z\d-]+\.){1,5}[a-z]{2,6}$/i
+
+export const validateFormInputs = (inputValues) => {
```
danielmajewski, line 4: this is validating specific form. Maybe rename to `validateSignUpFormInputs`

app/utils/validateFormInputs.js
```diff
@@ -0,0 +1,33 @@
...
+        errors.email = 'Email jest wymagany'
+    } else if (!emailPattern.test(email)) {
+        errors.email = 'Wprowadź prawidłowy adres email'
+    }
+
+    if (!emailConfirmation) {
+        errors.emailConfirmation = 'Potwierdzenie adresu email jest wymagane'
+    } else if (email !== emailConfirmation) {
+        errors.emailConfirmation = 'Adresy email muszą być takie same'
+    }
```
danielmajewski, line 30: You could extract these validators into seperate functions so someone can reuse then in different forms

e.g.:
```suggestion
const emailValidator = (value) => {
    if (!value) {    
        return 'Email jest wymagany'
    } else if (!emailPattern.test(email)) {
        return 'Wprowadź prawidłowy adres email'
    }
    
    return '';
}
```

So each validator returns an error or empty string (if there is no error);

app/components/Form/form-styling-components.js
```diff
@@ -0,0 +1,87 @@
...
+
+
+const FormFlex = styled.form`
+    display: flex;
+    flex-direction: column; 
+    max-width: 400px;
+`;
+
+export { FormWrapper, FormContent, Form, FormTitle, FormButton, FormRow, FormLabel, FormInput, ErrorMsg, Wrapper, FormFlex }
```
danielmajewski, line 87: why export like that? why not `export const FormFlex = styled.form...`

pages/user/sign-up.js
danielmajewski: So if form is valid (no errors) then you're clearing the form? Why?
I think you could add function that will contain all form data that would need to be send to API? Maybe `handleFormData`

PR #8, opened 2022-01-25
feat: sign up #P2022-361

rast-mk1:


pages/user/register.js
```diff
@@ -1,7 +1,147 @@
...
+        }
+        if (username.length < 5 && username.length > 0){
+            setUsernameError('Zbyt krótka nazwa użytkownika');
+            validated = false;
+        }
+        if (username.length > 20){
+            setUsernameError('Zbyt długa nazwa użytkownika');
+            validated = false;
+        }
+        let usernameRegEx = /[^A-Za-z0-9]+/;
```
MS151994, line 63: I would use "const"

rast-mk1: ok, const makes sense

pages/user/register.js
```diff
@@ -1,7 +1,147 @@
...
+        }
+        let emailRegEx = new RegExp('[a-z0-9]+@[a-z0-9]+[.]+[a-z]');
+        if (emailRegEx.test(mail) != true){
+            setMailError ('Niepoprawny format adresu e-mail');
+            validated = false;
+        } 
+        if (!mail) {
+            setMailError('Wprowadź adres e-mail');
+            validated = false;
+        }
```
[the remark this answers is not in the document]
rast-mk1: What do you mean exactly?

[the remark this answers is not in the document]
lenasokolova: I mean,
First, you're checking if email and email in confirm email input are matching.
Second, you're checking if the email format is correct using RegEx.
And third, you're checking if anyone at least enters any email in the input field.
I think the order has to be completely different. So you proly should first check if the input is not empty -> then RegEx -> and the last step is matching two emails.

Because as far as I know, Javascript statements are executed in the order they appear and it may be caused some problems in the future when your code is much longer and more complicated.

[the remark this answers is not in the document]
rast-mk1: Thank you. Your take on this is correct and seems more refined.

pages/user/register.js
```diff
@@ -1,7 +1,147 @@
...
+        setMailError('');
+
+        let validated = true;
+
+        // TODO sprawdź, czy taki użytkownik istnieje w users
+
+        // if (checkUserExists(username) === true){
+        //     alert('istnieje');
+        // }
+        // if (!checkUserExists(username) === false){
```
rast-mk1, line 47: note to self: get rid of !

marcelina-hasiak (commented):
Remember to delete unnecessary comments next time. Thanks 😊

pages/user/register.js
```diff
@@ -1,7 +1,147 @@
...
+//    return users;
+//}
+
+//const checkUserExists = async (username) => {
+//    const users = await getUsers()
+//    .then (result => {
+//        return result.some(user => user.username === username);
+//    });
+//    
+//}
```
MS151994, line 28: In my opinion, I would send a query to the backend and search "username" by name there, and return (true or false). and  displayed some message for user.

pages/user/register.js
```diff
@@ -1,7 +1,147 @@
...
+        }
+
+
+    }    
+    
   return (
-    <div >
-        user register
-    </div>
+    <form style={{margin: "auto", marginTop: "5%", display: "flex", flexDirection: "column", justifyContent: "center", width: "25%"}}>
```
weronika299, line 110: Maybe it's better to extract form to a separate styled-component

pages/user/register.js
```diff
@@ -1,7 +1,147 @@
...
+            value={mail}
+            error={mailError}
+            onInputChange={(e) => setMail(e.target.value)} />
+        <Input 
+            id="reMail" 
+            name="reMail" 
+            type="email" 
+            label="Confirm User e-mail:"
+            value={reMail}
+            // error={reMailError}
```
weronika299, line 142: unnecessary comment

app/components/Form/index.js
```diff
@@ -1,22 +1,21 @@
-import styled from 'styled-components';
+let value = "";
+const onInputChange = (newValue) => {
+    value = newValue;
+}
```
weronika299, line 4: I don't understand why this is needed

PR #9, opened 2022-01-26
feat: Sign up #P2022-366

MS151994:


pages/user/register.js
```diff
@@ -1,7 +1,109 @@
...
+      setUserPasswordError("password is required");
+    } else if (password.length < 8) {
+      setUserPasswordError("password should be a min 8 characters");
+    }
+
+    if (email === "") {
+      setUserEmailError("email is required!");
+    } else if (email.indexOf("@") === -1) {
+      setUserEmailError("check your email adress!");
+    }
```
lenasokolova, line 45: I guess the indexOf method is checking if there is any "@", and it will return the first appearance in a string, but if there are many of them it will omit the next instances, apart from that it is worth thinking about top-level domains checking (.com, .org. etc.)

pages/user/register.js
```diff
@@ -1,7 +1,109 @@
-export default function Home() {
+import { Input, FormFlex } from "components/Form/index";
+import { useState } from "react";
+import styled from "styled-components";
+
+const Wrapper = styled.div`
+  width: 50%;
+  margin: 50px auto;
+  background-color: ${(props) => (props.theme === "light" ? "#fff" : "#666")};
+`;
```
lenasokolova, line 9: it's better to have components in a separate file, then export them and import if you want to use it in a specific file, so you can use Wrapper in your sign up form as in log in form.

pages/user/register.js
```diff
@@ -1,7 +1,109 @@
...
+      setUserEmailError("email is required!");
+    } else if (email.indexOf("@") === -1) {
+      setUserEmailError("check your email adress!");
+    }
+
+    if (confirmEmail === "") {
+      setUserConfirmEmailError("email is required!");
+    } else if (confirmEmail.indexOf("@") === -1) {
+      setUserConfirmEmailError("check your email adress!");
+    }
```
rast-mk1, line 51: could use regular expression instead

pages/user/register.js
```diff
@@ -1,7 +1,109 @@
...
+          type="password"
+          label="Password"
+          value={password}
+          error={userPasswordError}
+          onInputChange={(e) => setPassword(e.target.value)}
+        />
+        <Input
+          id="email"
+          name="email"
+          type="text"
```
marcelina-hasiak, line 88: For inputs with email adress we should use type="email".

marcelina-hasiak (commented):
Try to separate your files and commit more next time.
Good job ;)

pages/user/register.js
```diff
@@ -1,7 +1,109 @@
...
+
+    if (email !== confirmEmail) {
+      setUserEmailError("email must be the same!");
+      setUserConfirmEmailError("email must be the same!");
+    }
+
+    if (username && password && email && confirmEmail) {
+      const newUser = { username, password, email };
+      console.log(newUser);
+    }
```
rast-mk1, line 61: Seems like this block of code can use user data that is not validated yet

MS151994: I only use it to avoid loading the user object on to the console.log with each save.,only after validation. Thanks :)

pages/user/register.js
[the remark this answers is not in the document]
MS151994: I agree with you, but I meant a specific indication to the user what must be the same. In the future, for example: underline the entire field. Thanks :)

pages/user/register.js
```diff
@@ -1,7 +1,109 @@
...
+          id="email"
+          name="email"
+          type="text"
+          label="Email"
+          value={email}
+          error={userEmailError}
+          onInputChange={(e) => setEmail(e.target.value)}
+        />
+        <Input
+          id="Confirmemail"
```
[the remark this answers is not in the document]
MS151994: Good point, thanks.

pages/user/register.js
MS151994: Good point, thanks. :)

pages/user/register.js
MS151994: Next time. I have to learn and understand this. :)

pages/user/register.js
MS151994: I agree with you, I will correct it :)

pages/user/register.js
MS151994: Yes, indexof returns the position of the found element, if it does not exist, returns: -1. Next time I'll try to use a regular expression

pages/user/register.js
```diff
@@ -1,7 +1,109 @@
...
+          name="Confirmemail"
+          type="text"
+          label="Confirmemail"
+          value={confirmEmail}
+          error={userConfirmEmailError}
+          onInputChange={(e) => setConfirmEmail(e.target.value)}
+        />
+        <button onClick={handleSubmit} type="submit">
+          Register
+        </button>
```
weronika299, line 105: Maybe it is worth separating the button component? What do you think?

MS151994: I forgot that we use this button in the login form, in fact it should be separated into a separate component. Thanks 🙂

pages/user/register.js
```diff
@@ -1,7 +1,242 @@
...
+      const newUser = {
+        username: values.username,
+        password: values.password,
+        email: values.email,
+      };
+      console.log(newUser);
+    } else {
+      setErrors({
+        ...errors,
+      });
```
danielmajewski, line 158:
```suggestion
            setErrors({
        ...validation.errors,
      });
```

pages/user/register.js
```diff
@@ -1,7 +1,242 @@
...
+    } else {
+      comparisonEmails = true;
+    }
+
+    // TODO nie działa ustawianie errora w tej funkcji.
+
+    if (username && password && email && confirmEmail && comparisonEmails) {
+      correct = true;
+    }
+    return { username, password, email, correct };
```
danielmajewski, line 200:
```suggestion
    return { username, password, email, correct, errors: {
      ...errors,
      emailError: errMessage.comparisonEmailsError,
      confirmEmailError: errMessage.comparisonEmailsError,
    } };
```

PR #16, opened 2022-02-02
feat: Add font awesome icon to project #P2022-610

rast-mk1:
To use Font Awesome icons:

first `import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';` in your page
and `import { faRuler } from '@fortawesome/free-solid-svg-icons';` specific icons

then `<FontAwesomeIcon icon={faRuler} />` use your icon


pages/_app.js
```diff
@@ -1,5 +1,8 @@
 import Layout from 'components/Layout';
 import '../styles/globals.css'
+import { config } from '@fortawesome/fontawesome-svg-core'
+import '@fortawesome/fontawesome-svg-core/styles.css'
+config.autoAddCss = false
```
danielmajewski, line 5: is it possible to make seperate file for this import and configuration? E.g. `fonts.js` and just `import '..../fonts', something like that?

rast-mk1: will look into this, thanks

pages/fontawesome.js
```diff
@@ -0,0 +1,3 @@
+import { config } from '@fortawesome/fontawesome-svg-core';
+import '@fortawesome/fontawesome-svg-core/styles.css';
+config.autoAddCss = false;
```
DianaSzalek, line 3: Please move all of this into `global.css` that shouldn't be a page

rast-mk1: Yes, I am now aware of that. Will fix it. Thank you!

PR #20, opened 2022-02-04
feat: footer component #613

weronika299:
Icons will be added in a separate subtask P2022-734, because I'm waiting for a ticket P2022-610 (Add font awesome icons to project).

Edit: I added icons because task P2022-610 is already on develop branch, so I had access to icons.


app/components/Footer/index.js
```diff
@@ -1,3 +1,72 @@
...
+                <LinkStyle><a>Media Center</a></LinkStyle>
+                <LinkStyle><a>Investor Relations</a></LinkStyle>
+                <LinkStyle><a>Jobs</a></LinkStyle>
+                <LinkStyle><a>Terms of Use</a></LinkStyle>
+                <LinkStyle><a>Privacy</a></LinkStyle>
+                <LinkStyle><a>Legal Notices</a></LinkStyle>
+                <LinkStyle><a>Cookie Preferences</a></LinkStyle>
+                <LinkStyle><a>Corporate Information</a></LinkStyle>
+                <LinkStyle><a>Contact Us</a></LinkStyle>
+            </LinksList>
```
marcelina-hasiak, line 26: It's quite a long list of <LinkStyle> repeated component. According to DRY principle, perhaps it would be better to gather all the data connected to links in one object/array and then iterate through  it. Check it out: https://reactjs.org/docs/lists-and-keys.html

app/components/Footer/index.js
```diff
@@ -1,3 +1,72 @@
...
+        column-gap: 3rem;
+    }
+`;
+
+export const LinkStyle = styled.li`
+    cursor: pointer;
+    &:hover {
+        text-decoration: underline;
+    }
+`;
```
marcelina-hasiak, line 72: I had the same in my previous PR. Just add an empty line (+73) ;)

app/components/Footer/index.js
```diff
@@ -1,3 +1,72 @@
...
+        </FooterWrapper>
+    );
 }
+
+export const FooterWrapper = styled.div`
+    display: grid;
+    justify-content: center;
+    bottom: 0;
+    width: 100%;
+    color: grey;
```
marcelina-hasiak, line 37: Isn't a better way to declare colors with rgb or hex code? If they're the same as on Netflix site it'd be easier to replace them across the whole app when we get final designs.

app/components/Footer/index.js
```diff
@@ -1,3 +1,72 @@
...
+            </LinksList>
+            <p style={{fontSize: "0.8rem", marginLeft: "2.40rem", marginTop:"2rem"}}>© 2022 InTiVi</p>
+        </FooterWrapper>
+    );
 }
+
+export const FooterWrapper = styled.div`
+    display: grid;
+    justify-content: center;
+    bottom: 0;
```
marcelina-hasiak, line 35: Default `position` property is `static`, so `bottom: 0` does nothing. 
To put footer exactly on the bottom of the page some styling property should be added also to its parent.

app/components/Footer/index.js
```diff
@@ -1,3 +1,72 @@
...
+                <LinkStyle><a>Cookie Preferences</a></LinkStyle>
+                <LinkStyle><a>Corporate Information</a></LinkStyle>
+                <LinkStyle><a>Contact Us</a></LinkStyle>
+            </LinksList>
+            <p style={{fontSize: "0.8rem", marginLeft: "2.40rem", marginTop:"2rem"}}>© 2022 InTiVi</p>
+        </FooterWrapper>
+    );
 }
+
+export const FooterWrapper = styled.div`
```
marcelina-hasiak, line 32: By declaring `FooterWrapper` as div we've lost semantic tag which is `footer`

app/components/Footer/index.js
```diff
@@ -1,3 +1,77 @@
...
+    list-style-type: none;
+`;
+
+export const LinksList = styled.ul`
+    display: grid;
+    grid-template-columns: 1fr 1fr 1fr 1fr;
+    column-gap: 7rem;
+    row-gap: 1.1rem;
+    list-style-type: none;
+    @media (max-width: 992px) {
```
danielmajewski, line 59: can you put resolution breakpoint into separate file?

app/components/Footer/index.js
```diff
@@ -1,3 +1,77 @@
...
+        </FooterWrapper>
+    );
+}
+
+export const FooterWrapper = styled.footer`
+    display: grid;
+    justify-content: center;
+    width: 100%;
+    font-weight: 600;
+    color: rgb(94, 93, 93);
```
danielmajewski, line 42: I think that HEX colors are enough for us.

app/components/Footer/index.js
```diff
@@ -1,3 +1,77 @@
...
 }
+
+function List(props) {
+    const links = props.links;
+    const listItems = links.map((link, index) =>
+    <Item key={index} value={link} />
+    );
+    return (
+        <LinksList>
+        {listItems}
```
danielmajewski, line 17: indent

app/components/Footer/index.js
```diff
@@ -1,3 +1,77 @@
+import styled from "styled-components";
```
danielmajewski, line 1: use ' instead of "

app/components/Footer/index.js
```diff
@@ -1,3 +1,77 @@
...
+
 export default function Footer() {
-    return <footer />;
+const links = ['Audio and Subtitles', 'Audio Description', 'Help Center', 'Gift Cards', 'Media Center', 'Investor Relations', 'Jobs', 'Terms of Use', 'Privacy', 'Legal Notices', 'Cookie Preferences', 'Corporate Information', 'Contact Us'];
```
danielmajewski, line 4:
```suggestion
const mockLinks = [
    'Audio and Subtitles',
    'Audio Description',
    'Help Center',
    ...
```

app/components/Footer/index.js
```diff
@@ -1,3 +1,104 @@
...
+    'Contact Us'];
+
+  function Item(props) {
+    return <ListItem><a>{props.value}</a></ListItem>;
+  }
+
+  function List(props) {
+    const links = props.links;
+    const listItems = links.map((link, index) =>
+      <Item key={index} value={link} />
```
wlechowicz, line 29: Using array index as `key` is and anti-pattern. You can read more [here](https://pl.reactjs.org/docs/lists-and-keys.html#keys) and [here](https://medium.com/wesionary-team/using-index-as-a-key-is-an-anti-pattern-in-react-8e5db3aea3a9). Your `mockLinks` should be a list of objects, each containing a key, a display string and an href at minimum. If you don't want to play around with fake hrefs, then don't use `<a>` at all, just use a `<span>` with styling. An `<a>` without `href` is a no-no.

app/components/Footer/index.js
```diff
@@ -1,3 +1,104 @@
...
+    );
+  }
+
+  return (
+    <FooterWrapper>
+      <IconWrapper>
+        <FontAwesomeItem> <FontAwesomeIcon icon= { faFacebookF } /> </FontAwesomeItem>
+        <FontAwesomeItem> <FontAwesomeIcon icon= { faInstagram } /> </FontAwesomeItem>
+        <FontAwesomeItem> <FontAwesomeIcon icon= { faTwitter } /> </FontAwesomeItem>
+        <FontAwesomeItem> <FontAwesomeIcon icon= { faYoutube } /> </FontAwesomeItem>
```
wlechowicz, line 44: That's a lot of the same code repeated, enough to make it a small local component that takes one prop and saves you from the copy and paste.

app/components/Footer/breakpoints.js
```diff
@@ -0,0 +1,11 @@
...
+	mobileL: '425px',
+	tablet: '768px',
+	desktop: '1024px'
+};
+
+export const device = {
+	mobileL: `(min-width: ${size.mobileL})`,
+	tablet: `(min-width: ${size.tablet})`,
+	desktop: `(min-width: ${size.desktop})`
+};
```
wlechowicz, line 11: These should be common for the whole app, not footer specific. Also the 425 breakpoint seems unnecessary. Tablet should be the only one we need for mobile-first approach.

app/components/Footer/index.js
```diff
@@ -1,3 +1,104 @@
...
+      </IconWrapper>
+      <List links={mockLinks} />
+      <p style={{ fontSize: "0.8rem", marginLeft: "2.40rem", marginTop: "2rem" }}>© 2022 InTiVi</p>
```
wlechowicz, line 47: Make this a styled component.

app/components/Footer/index.js
```diff
@@ -1,3 +1,104 @@
...
+  display: flex;
+  margin-left: 0.4rem;
+  margin-top: 4rem;
+  margin-bottom: 0.5rem;
+  font-size: 1.5rem;
+  color: #FFFFFF;
+  list-style-type: none;
+`;
+
+export const FontAwesomeItem = styled.li`
```
wlechowicz, line 74: Doesn't seem like anything about this component makes it Font Awesome specific, so a more generic name would be better, like IconWrapper.

app/components/Footer/index.js
```diff
@@ -1,3 +1,104 @@
...
+  width: 100%;
+  font-weight: 600;
+  color: #5e5d5d;
+  background-color: #000000;
+  @media ${device.tablet} {
+    justify-content: center;
+  }
+`;
+
+export const IconWrapper = styled.ul`
```
wlechowicz, line 64: `IconList` would make more sense, since it's a list.

wlechowicz (changes requested):
Prettier will fix the formatting issues automatically, so not commenting on those.

app/components/Footer/index.js
```diff
@@ -1,3 +1,151 @@
...
+    column-gap: 2.3rem;
+  }
+`;
+
+export const LinkItem = styled.li`
+  cursor: pointer;
+  &:hover {
+    text-decoration: underline;
+  }
+`;
```
rast-mk1, line 145: Could you make it that not the entire list item, but rather just the caption inside it is underlined and with cursor: pointer on hover?

rast-mk1: Sure, that's also fine.

app/components/Footer/index.js
```diff
@@ -1,3 +1,151 @@
...
+    },
+    {
+      id: 12,
+      displayLink: 'Corporate Information',
+    },
+    {
+      id: 13,
+      displayLink: 'Contact Us',
+    },
+  ];
```
hiimsagan, line 59: Perhaps i will be better to have that in other file ? @wlechowicz @danielmajewski

hiimsagan: Same think to style components ?

weronika299: Ok, I will fix it. Should I make the same with styled components?

wlechowicz: Yes, commented on that in my review.

weronika299: Ok, I will correct it

pages/index.js
```diff
@@ -18,8 +18,6 @@ export default function Home() {
           Please <Link href={'/user/login'}>sign in</Link>.
         </p>
       </main>
-
-      <footer className={styles.footer}>(c) intive 2022</footer>
```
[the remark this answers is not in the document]
wlechowicz: Never mind, it's used through the Layout component!

hiimsagan (approved):
Well done

wlechowicz (approved):

wlechowicz:
LGTM

PR #21, opened 2022-02-04
feat: #P2022-610 Add font awesome icon to project

rast-mk1:
I see that to show this Icon you have to:

import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
import { faCoffee } from '@fortawesome/free-solid-svg-icons'

and inside component:
<FontAwesomeIcon icon={faCoffee} />


styles/fontawesome.js
```diff
@@ -0,0 +1,3 @@
+import { config } from '@fortawesome/fontawesome-svg-core';
```
danielmajewski, line 1: where are you importing this file so styles are enabled?

rast-mk1: Let's say that you would like to use icons in `pages/index.js` - then you `import '../styles/fontawesome.js';` inside `pages/index.js`

rast-mk1:
`import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';`
then:
`import { faRuler } from '@fortawesome/free-solid-svg-icons';`
or
`import { faBookmark } from '@fortawesome/free-regular-svg-icons';`
or
`import { faYoutube } from '@fortawesome/free-brands-svg-icons';`

and use icon: `<FontAwesomeIcon icon={faRuler} /> `

for styling icons, one way to do that is to give your `<FontAwesomeIcon />`  some `id` and style it inside `globals.css`

PR #22, opened 2022-02-06
feat: featured video section #611

wolnio:


wolnio:
URL addres for this feature: /banner/video-section

pages/banner/video-section.js
```diff
@@ -0,0 +1,15 @@
+import { Wrapper, Contents, Title, Media } from "components/Banner";
+
+export default function VideoSection() {
+  const title = "Whiplash";
+  return (
+    <>
+      <Wrapper imgLink={"https://wallpaperaccess.com/full/2191678.jpg"}>
```
danielmajewski, line 7: can you download this image as our internal not external source? So we could just import it inside our project?

app/components/Banner/index.js
```diff
@@ -0,0 +1,80 @@
...
+export function Media() {
+  const [isMute, setMute] = useState("Mute");
+
+  const toggleMute = () => {
+    if (isMute === "Mute") {
+      setMute("Unmute");
+    } else if (isMute === "Unmute") {
+      setMute("Mute");
+    }
+  };
```
danielmajewski, line 72: isMute should anser only Yes or No, so use as boolean values only if prefix `is` for variables.

```suggestion
  const [isMute, setMute] = useState(true);

  const toggleMute = () => setMute(!isMute);
```

wlechowicz:
Not sure why checks aren't running on this, I reverted my dummy commits that were supposed to trigger them.

wlechowicz:
Oh I know, @wolnio please pull in develop onto this branch and push the merged changes to origin.

pages/banner/video-section.js
```diff
@@ -0,0 +1,15 @@
+import { Wrapper, Contents, Title, Media } from "components/Banner";
+
+export default function VideoSection() {
+  const title = "Whiplash";
+  return (
+    <>
+      <Wrapper imgLink={'https://wallpaperaccess.com/full/2191678.jpg'}>
+        <Contents flexDir={"column"}>
```
wlechowicz, line 8: Does it make sense to have the flex-direction configurable like this? Setting it to `row` doesn't seem to work well anyway, since title is touching the button.

pages/banner/video-section.js
```diff
@@ -0,0 +1,15 @@
+import { Wrapper, Contents, Title, Media } from "components/Banner";
```
wlechowicz, line 1: This should be just one component that takes a movie object through props and does the rest internally. If I want to use a movie hero component, I want just one component, I don't want to spell out the whole structure inside (Contents, Title, Media).

wlechowicz (changes requested):
1. Rename the component from Banner to MovieHero or FeaturedVideo, Banner is misleading
2. Split up the components into separate files, for example `app/components/MovieHero/MuteIcon.js`, `MovieHero/PlayIcon.js` etc which get imported by the main component in index.js. These icons will likely get reused in the future so might get extracted into common components later.

wlechowicz (approved):

wlechowicz:
LGTM

PR #26, opened 2022-02-08
feat: add SimpleHeader component

marcelina-hasiak:


app/components/Header/simple-header.js
```diff
@@ -0,0 +1,37 @@
...
+  );
+};
+
+export default SimpleHeader;
+
+const HeaderContainer = styled.header`
+  display: flex;
+  justify-content: center;
+  padding: 2rem 1rem 0;
+  @media only screen and (min-width: 768px) {
```
wlechowicz, line 26: Please use this approach to define reusable media queries: https://jsramblings.com/how-to-use-media-queries-with-styled-components/
If we're going with mobile-first approach, then we only need 1 breakpoint: `tablet: 768px` so that base CSS is for mobile up to 768 and then inside media query it's for tablets and desktops 768 and up. (like you did here)

app/components/Header/simple-header.js
```diff
@@ -0,0 +1,37 @@
+import styled from "styled-components";
+import Link from "next/link";
+
+import LogoLink from "components/LogoLink/logo-link";
```
wlechowicz, line 4: This is an unnecessary nesting in import path caused by misnaming of the module. Module should be named 'components/LogoLink/index.js' or 'components/LogoLink.js', in both cases the import becomes 'from "components/LogoLink"'.
Same applies to this Header component.

app/components/Header/simple-header.js
```diff
@@ -0,0 +1,37 @@
...
+import { PrimaryButton } from "../Button/button";
```
wlechowicz, line 5: 1. Button button, ditto
2. Import path should be "components/Button"

wlechowicz (changes requested):

marcelina-hasiak:
Attention: I'm currently waiting for this PR: https://github.com/intive/patronage22_poznan_js/pull/29 to be resolved before this one.

app/components/Header/simple-header.js
marcelina-hasiak: Thank you for this neat pattern ;)

app/components/SimpleHeader/index.js
```diff
@@ -0,0 +1,39 @@
+import styled from "styled-components";
+import Link from "next/link";
+
+import { device } from "../../../styles/breakpoints";
+import LogoLink from "../LogoLink/index";
+import PrimaryButton from "../PrimaryButton/index";
```
wlechowicz, line 6: Import path for components should be `components/` and you don't need to add `index` at the end, so: `from "components/PrimaryButton";`
I assume the double quotes are here because you didn't apply Prettier to this PR yet.

wlechowicz (changes requested):

wlechowicz:
> For me it looks good.
@weronika299 then you should add your review and choose "Approve".

weronika299 (approved):

marcelina-hasiak:
Zrzut ekranu 2022-02-11 o 22 30 49

wlechowicz (approved):

wlechowicz:
LGTM

PR #27, opened 2022-02-09
feat: search functionality #674 

MS151994:
Please check and make suggestions for the component. I am not sure if this is how it should be in terms of code.


wlechowicz:
Feature branch not started from develop, reopen from a fixed base branch.

PR #29, opened 2022-02-09
feat: PrimaryButton component

marcelina-hasiak:


app/components/PrimaryButton/index.js
```diff
@@ -0,0 +1,12 @@
+import styled from "styled-components";
+
+export const PrimaryButton = styled.button`
```
wlechowicz, line 3: If there's only one export in a module, it should be a default export.

wlechowicz (changes requested):

app/components/PrimaryButton/index.js
```diff
@@ -0,0 +1,14 @@
+import styled from 'styled-components';
+
+const PrimaryButton = styled.button`
```
danielmajewski, line 3: 1. Why not input type button?
2. primary button suggest that there will be secondary and so on... Can you put it in directory like `components/buttons/primaryButton.js ...

wlechowicz (approved):

app/components/PrimaryButton/index.js
wlechowicz: @danielmajewski can you elaborate on 1.? I'm not sure I get what you mean. 
As for 2, term "primary" just means the main call to action button, but it's not a bad idea to have components/Button/index.js and PrimaryButton as one of the exports.

marcelina-hasiak: @danielmajewski This button will be used as Sign In button as well as Link to SignIn page.

PR #30, opened 2022-02-09
feat: search funcionality #674

MS151994:


app/components/Header/Search.js
```diff
@@ -0,0 +1,35 @@
...
+import { SearchMoviesContainer, SearchButton, SearchInput, ClearInputButton } from './searchStyled';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSearch } from '@fortawesome/free-solid-svg-icons';
+
+export default function SearchMoviesInput() {
+  const [buttonState, setButtonState] = useState(false);
+  const [searchInputValue, setSearchInputValue] = useState('');
+
+  const searchIcon = <FontAwesomeIcon icon={faSearch} />;
+  const closeIcon = 'X';
```
rast-mk1, line 11: `faXmark `will be provided with  chore: update font awesome packages #32

app/components/Header/Search.js
```diff
@@ -0,0 +1,35 @@
...
+      {buttonState && (
+        <>
+          <SearchButton onClick={() => setButtonState(!buttonState)}>{searchIcon}</SearchButton>
+          <SearchInput
+            placeholder="Title, Actors, Film genre"
+            value={searchInputValue}
+            onChange={(e) => setSearchInputValue(e.target.value)}
+          ></SearchInput>
+          {searchInputValue.length > 0 ? (
+            <ClearInputButton onClick={() => setSearchInputValue(' ')}>
```
wolnio, line 24: When the field is cleared, its value is set to `' '` (1 space) so user can't see the placeholder anymore, until he presses backspace. Also after cleaning if user starts typing again, value prop contains 1 space at the beginning.

app/components/Header/Search.js
wlechowicz: A symmetrical X mark can be achieved with `&times;` which is a multiplication character: &times; which can be scaled with font-size, colored etc.

pages/api/movie/[id].js
```diff
@@ -0,0 +1,7 @@
+const movies = require('/public/shared/jsons/movieMockData.json');
+
+export default async function handler(req, res) {
+  const { id } = req.query;
+  const movieSearchResult = movies.find((movie) => movie.id === Number(id));
+  res.status(200).json(movieSearchResult || {});
```
Comment (wlechowicz) [inline pages/api/movie/[id].js:6]: Since we're searching for a specific movie by Id, the response should be HTTP 404 Not Found if not found.

.prettierrc.json
```diff
@@ -1,5 +1,4 @@
 {
   "printWidth": 100,
-  "singleQuote": true,
-  "endOfLine": "auto"
+  "singleQuote": true
 }
```
hiimsagan, line 4: These changes are nesseery ? @wlechowicz

package.json
```diff
@@ -9,10 +9,10 @@
...
   },
   "dependencies": {
-    "@fortawesome/fontawesome-svg-core": "^1.3.0",
-    "@fortawesome/free-brands-svg-icons": "^6.0.0",
-    "@fortawesome/free-regular-svg-icons": "^6.0.0",
-    "@fortawesome/free-solid-svg-icons": "^6.0.0",
+    "@fortawesome/fontawesome-svg-core": "^1.2.36",
+    "@fortawesome/free-brands-svg-icons": "^5.15.4",
+    "@fortawesome/free-regular-svg-icons": "^5.15.4",
+    "@fortawesome/free-solid-svg-icons": "^5.15.4",
```
rast-mk1, line 15: You should have versions from develop.

PR #32, opened 2022-02-09
chore: update font awesome packages

rast-mk1:
This will provide more icons since Font Awesome 6 was released recently


wlechowicz:
As discussed on daily call, we do not want to use unstable (beta) packages.

PR #40, opened 2022-02-16
feat: login page component

weronika299:


app/components/Button/index.js
```diff
@@ -2,8 +2,8 @@ import PropTypes from 'prop-types';
 import StyledButton from './Button.styles';
 import Link from 'next/link';
 
-const Button = ({ children, href, ...props }) => (
-  <>
+const Button = ({ children, href, className, ...props }) => (
+  <div className={className}>
```
wlechowicz, line 6: This `div` can break layout, move the `className` to Link for href variant and to StyledButton otherwise.

app/components/Form/Input/index.js
```diff
@@ -1,12 +1,14 @@
 import PropTypes from 'prop-types';
+import StyledInput from './Input.styles';
+import { ErrorMessage } from './Input.styles';
 
-const Input = ({ id, name, label, value, onInputChange, error, ...props }) => {
+const Input = ({ id, name, label, value, onInputChange, error, className, ...props }) => {
   return (
-    <>
+    <div className={className}>
```
wlechowicz, line 7: Same issue a with Button, `div` makes it a block.

app/components/LoginPage/LoginInputStyles.js
```diff
@@ -0,0 +1,14 @@
...
+
+export const LoginInput = styled(Input)`
+  input {
+    width: 100%;
+    margin-bottom: 1rem;
+    padding: 0.7rem 0.5rem;
+    letter-spacing: 0.3px;
+  }
+  label {
+    display: none;
```
wlechowicz, line 12: Add a prop to Input that removes the label via conditional rendering. But it has to add an `aria-label` attribute to the input instead.

wlechowicz (changes requested):

wlechowicz (approved):

wlechowicz:
LGTM!

PR #41, opened 2022-02-17
feat: adjust header to user authentication state

weronika299:


app/components/Layout/index.js
```diff
@@ -1,10 +1,14 @@
...
 
 export default function Layout({ children }) {
+  const { user, headerMode } = useContext(AppContext);
   return (
     <>
-      <Header />
+      {user ? <Header /> : <SimpleHeader headerMode={headerMode}></SimpleHeader>}
       {children}
       <Footer />
     </>
```
hiimsagan, line 14: Why do you doing changes here if you do that in previous branch feat: login page component #40

hiimsagan (changes requested):
Why you duplicate previous branch ? I think you will have conflicts here

app/components/Layout/index.js
weronika299: I separated the changes regarding the header to another pull request for better readability and it is not directly related to the login page.

app/components/Layout/index.js
```diff
@@ -1,10 +1,14 @@
 import Footer from '../Footer';
 import Header from '../Header';
+import { useContext } from 'react';
+import SimpleHeader from 'components/SimpleHeader';
+import AppContext from 'context/app';
 
 export default function Layout({ children }) {
+  const { user, headerMode } = useContext(AppContext);
```
wlechowicz, line 8: Move `headerMode` from Layout to SimpleHeader since Layout doesn't need this prop for anything. Let SimpleHeader read that value since it needs it.

wlechowicz (changes requested):

wlechowicz (approved):

app/components/Layout/index.js
```diff
@@ -1,10 +1,23 @@
 import Footer from '../Footer';
 import Header from '../Header';
+import Button from '../Button/Button.styles';
+import SimpleHeader from 'components/SimpleHeader';
+
+import { useActions, logIn, logOut } from 'actions';
```
wlechowicz, line 6: Rename your `app/actions/index.js` file to `app/actions/user.js` ad then adjust this import.

app/actions/index.js
```diff
@@ -0,0 +1,45 @@
...
+export const useActions = (initialState) => {
+  const [state, newDispatch] = useReducer(appReducer, initialState);
+  actions.dispatch = newDispatch;
+  return state;
+};
+
+// actions
+
+export const logIn = () => {
+  actions.dispatch({ type: actions.LOG_IN });
```
wlechowicz, line 40: You forgot to pass in the payload - the `user` object. This is working only because you have a bug, the user state is undefined initially (because you don't pass initial state in `const user = useActions();`) and then LOG_IN changes it from undefined to empty object.

app/components/Layout/index.js
```diff
@@ -1,10 +1,23 @@
...
 
 export default function Layout({ children }) {
+  const user = useActions();
```
wlechowicz, line 9: Missing initial state (an empty object in this case).

app/components/Layout/index.js
```diff
@@ -1,10 +1,23 @@
...
   return (
     <>
-      <Header />
+      {user ? <Header /> : <SimpleHeader></SimpleHeader>}
```
wlechowicz, line 12: After you fix the initial state and LOG_IN payload, this condition needs to change into something like `user.username` instead of just `user`, because that will be always truthy.

wlechowicz (changes requested):

app/actions/user.js
```diff
@@ -0,0 +1,46 @@
...
+export const useActions = (initialState) => {
+  const [state, newDispatch] = useReducer(appReducer, initialState);
+  actions.dispatch = newDispatch;
+  return state;
+};
+
+// actions
+
+export const logIn = (user) => {
+  actions.dispatch({ type: actions.LOG_IN, user });
```
wlechowicz, line 41: Action should have `type` and `payload` fields always.

app/actions/user.js
```diff
@@ -0,0 +1,46 @@
...
+};
+
+export const logOut = (user) => {
+  actions.dispatch({ type: actions.LOG_OUT, user });
```
wlechowicz, line 45: logOut doesn't need a payload.

app/actions/user.js
```diff
@@ -0,0 +1,46 @@
...
+  LOG_OUT: 'log-out-action',
+  dispatch: (action) => {
+    console.error('Actions was used before dispatch was ready - app store', action);
+  },
+};
+
+// reducer
+
+export const appReducer = (state, action = {}) => {
+  const { type, user } = action;
```
wlechowicz, line 14: `type` and `payload`.

app/actions/user.js
```diff
@@ -0,0 +1,46 @@
...
+
+  switch (type) {
+    case actions.LOG_IN:
+      return { ...user };
+
+    case actions.LOG_OUT:
+      return { ...user };
```
wlechowicz, line 21: Should return an empty object instead `{}`

app/components/Layout/index.js
```diff
@@ -1,10 +1,23 @@
...
+  const user = useActions({});
   return (
     <>
-      <Header />
+      {user.username ? <Header /> : <SimpleHeader></SimpleHeader>}
+      <div style={{ position: 'absolute', bottom: '0', right: '0' }}>
+        <Button primary value="Log In" onClick={() => logIn({ username: 'Grażyna' })}>
+          Log In
+        </Button>
+        <Button primary value="Log Out" onClick={() => logOut({})}>
```
wlechowicz, line 17: No payload needed in logOut, reducer will know what to do.

app/components/Layout/index.js
```diff
@@ -1,10 +1,23 @@
...
+
+import { useActions, logIn, logOut } from 'actions/user';
 
 export default function Layout({ children }) {
+  const user = useActions({});
   return (
     <>
-      <Header />
+      {user.username ? <Header /> : <SimpleHeader></SimpleHeader>}
+      <div style={{ position: 'absolute', bottom: '0', right: '0' }}>
```
wlechowicz, line 13: Please add a comment that this is a temporary solution for development purposes only.

wlechowicz (changes requested):
Few minor fixes and good to go

wlechowicz (approved):

wlechowicz:
LGTM

PR #47, opened 2022-02-20
fix: layout of LogoLink component

wlechowicz:
Fixes the layout of LogoLink component which was not adjusting to its parent size.

### DIFF
--- a/app/components/Header/headerStyle.js
+++ b/app/components/Header/headerStyle.js
@@ -9,7 +9,7 @@ export const NavigationBar = styled.header`
   display: flex;
   justify-content: space-between;
   align-items: center;
-  padding: 1rem;
+  padding: 1.5rem 1rem;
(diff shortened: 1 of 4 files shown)


marcelina-hasiak (approved):

wlechowicz:
Issue of vertical sizing to be fixed by @marcelina-hasiak

marcelina-hasiak:
Zrzut ekranu 2022-02-22 o 00 30 25
Zrzut ekranu 2022-02-22 o 00 32 01
Zrzut ekranu 2022-02-22 o 00 31 37
Everything seems to work fine now, but the problem is I used `!important` rule to overwrite next/image styling => mainly position:absolute. Previously I didn't use '!mportant at all but I had to deal with image and span covering whole Header and cover its element (can't activate buttons/links).

PR #53, opened 2022-02-24
feat: footer storybook

weronika299:


wlechowicz (approved):
The only thing I notice is that the LinkList seems to be missing some padding on the right side. Also noticeable in the Footer. Other than that - great!

obraz

PR #55, opened 2022-02-26
refactor: replace FontAwesome with Icon everywhere

rast-mk1:


app/components/Icon/index.js
```diff
@@ -12,6 +12,7 @@ import {
   faChevronRight,
   faChevronDown,
   faGear,
+  faSortDown,
 } from '@fortawesome/free-solid-svg-icons';
 import { faFacebookF, faInstagram, faTwitter, faYoutube } from '@fortawesome/free-brands-svg-icons';
 const iconMap = {
```
weronika299, line 18: Why do you have declared Icons sometimes as string and sometimes not?

rast-mk1: FontAwesome icons are imported in this file and then "translated" to string inside `iconMap`. So `<FontAwesome icon={faTwitter} />` is basically the same as `<Icon type="twitter" />` and importing `faTwitter` takes place only in `Icon/index.js`.

rast-mk1: Ah, yes: Prettier changes single words like that and leaves multiple-word names inside quotes :)

hiimsagan (approved):
Looks good

app/components/Icon/index.js
weronika299: Aaa ok :) But is it work?  I ask because in mockLists file you have it inside quotes. :)

rast-mk1: Yes, it works fine.

weronika299: Ok, great :D

weronika299 (approved):

PR #56, opened 2022-02-27
feat:  loginPage (LoginForm) storybook

weronika299:


app/components/Form/LoginForm/index.js
```diff
@@ -0,0 +1,70 @@
...
+  CheckboxWrapper,
+  SignUpLinkParagraph,
+  SignUpLink,
+  CheckboxContainer,
+  SignInButton,
+} from './LoginForm.styles';
+
+import { LoginInput } from './LoginInput.styles';
+
+export default function LoginForm({ ...props }) {
```
wolnio, line 15: Is `{...props}` necessary, since you're not passing any prop in `LoginForm` in LoginPage/index.js ?

app/components/Form/LoginForm/index.js
```diff
@@ -0,0 +1,70 @@
...
+        error={usernameError}
+        onInputChange={(e) => setUsername(e.target.value)}
+      />
+      <LoginInput
+        id="password"
+        type="password"
+        placeholder="Password"
+        value={password}
+        onInputChange={(e) => setPassword(e.target.value)}
+      />
```
wolnio, line 47: Isn't there `passwordError` message missing? Shouldn's we inform user if password is incorrect or the field is blank?

app/components/Form/LoginForm/index.js
weronika299: I need this because when I delete the {... props} the storybook will stop working.

app/components/Form/LoginForm/index.js
weronika299: Yes you are right, but I created only the page layout, not validation. Validation will be in a separate task.

app/components/Form/LoginForm/LoginForm.stories.js
```diff
@@ -0,0 +1,27 @@
...
+  title: 'Form/LoginForm',
+  component: LoginForm,
+};
+
+const Template = ({ ...args }) => <LoginForm {...args} />;
+
+export const DesktopLoginForm = Template.bind({});
+DesktopLoginForm.args = {
+  backgroundColor: 'rgb(0, 0, 0, 0.5)',
+  color: 'color: rgb(255, 255, 255);',
```
rast-mk1, line 13: Why does line `color: 'color: rgb(255, 255, 255);',` has `color:` twice?

weronika299: Yees, thanks, it was mistake. I have already fixed it.

rast-mk1 (approved):

PR #60, opened 2022-03-03
feat: redesign authentication page container

weronika299:


app/components/Pages/AuthenticationPageContainer/index.js
```diff
@@ -0,0 +1,17 @@
+import AuthenticationFormHeader from 'components/Form/AuthenticationFormHeader';
+import { Wrapper, FormContainer, Form } from './AuthenticationPageContainer.styles';
+
+export default function AuthenticationPageContainer(pageLayout = {}) {
+  const isSignInPage = pageLayout.signIn !== false;
```
marcelina-hasiak, line 5: In this situation we pass boolean values inside getServerSideProps() so simply `const isSignInPage = pageLayout.signIn` is all we need here.  Component can be extended with Prop types to secure ourselves that we expect boolean values

app/components/Form/AuthenticationFormHeader/index.js
```diff
@@ -0,0 +1,15 @@
+import { TabWrapper, FormHeader, Underlined } from './AuthenticationFormHeader.styles.js';
+
+export default function AuthenticationFormHeader(pageLayout = {}) {
+  const isSignInPage = pageLayout.signIn !== false;
+  return (
+    <FormHeader>
+      <h1>inTiVi</h1>
+      <TabWrapper>
+        <a href="sign-in">Sign in</a>
+        <a href="create-account">Sign up</a>
```
marcelina-hasiak, line 10: use `<Link>` in such situations

app/components/Form/AuthenticationFormHeader/index.js
```diff
@@ -0,0 +1,15 @@
...
+      </TabWrapper>
+      {isSignInPage ? <Underlined /> : <Underlined style={{ marginLeft: '102px' }} />}
```
marcelina-hasiak, line 12: I don't think we need separate component just to underline active form, you can do it with props and css only

marcelina-hasiak (changes requested):

app/components/Form/AuthenticationFormHeader/index.js
```diff
@@ -0,0 +1,15 @@
+import { TabWrapper, FormHeader, Underlined } from './AuthenticationFormHeader.styles.js';
+
+export default function AuthenticationFormHeader(pageLayout = {}) {
+  const isSignInPage = pageLayout.signIn !== false;
+  return (
+    <FormHeader>
+      <h1>inTiVi</h1>
```
marcelina-hasiak, line 7: Logo is ready to be applied here.

app/components/Pages/AuthenticationPageContainer/AuthenticationPageContainer.styles.js
```diff
@@ -0,0 +1,44 @@
...
+export const Wrapper = styled.div`
+  display: flex;
+  height: 100vh;
+  width: 100vw;
+  background-color: #2a2535;
+  color: white;
+  ${device.tablet} {
+    justify-content: flex-end;
+    align-items: center;
+    background-image: url('../images/background-image.jpg');
```
marcelina-hasiak, line 12: What do you think about giving our background image more semantic name? It could be more background images in the future, so marking this particular image as hero one would be a good move.

pages/user/sign-in.js
```diff
@@ -0,0 +1,10 @@
+import AuthenticationPageContainer from 'components/Pages/AuthenticationPageContainer';
+
+export default function UserLogin() {
+  return <AuthenticationPageContainer />;
+}
+export async function getServerSideProps() {
+  return {
+    props: { pageLayout: { header: false, footer: false, signIn: true } },
+  };
+}
```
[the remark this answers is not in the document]
weronika299: Ok, right. I will fix it and I will make only one page.

app/components/Pages/AuthenticationPageContainer/AuthenticationPageContainer.stories.js
```diff
@@ -0,0 +1,24 @@
...
+
+export default {
+  title: 'Pages/AuthenticationPageContainer',
+  component: AuthenticationPageContainer,
+};
+
+const Template = (...args) => <AuthenticationPageContainer {...args} />;
+
+export const DesktopView = Template.bind({});
+
```
lenasokolova, line 11: don't you think you could use here `defaultViewport: 'responsive'`
?
like this:

```suggestion
export const DesktopView = Template.bind({});
DesktopView.parameters = {
  viewport: {
    defaultViewport: 'responsive',
  },
};
```

app/components/Form/AuthenticationFormHeader/index.js
wlechowicz: And an absolute path like "/sign-in" and "/sign-up".

app/components/Form/AuthenticationFormHeader/index.js
wlechowicz: A prop like `active` true/false that's true on active tab and the tab styles itself.

app/components/Pages/AuthenticationPageContainer/index.js
```diff
@@ -0,0 +1,20 @@
+import AuthenticationFormHeader from 'components/Form/AuthenticationFormHeader';
+import { Wrapper, FormContainer, Form } from './AuthenticationPageContainer.styles';
+import { useState } from 'react';
+
+export default function AuthenticationPageContainer({ active }) {
+  const [activeMode] = useState({ active });
```
wlechowicz, line 6: Seems to me like you don't need to store this `active` prop in state since you're not modifying it. You can just use the prop directly in render.

wlechowicz (approved):
Looks good, one change requested but overall approved.

app/components/Pages/AuthenticationPageContainer/index.js
weronika299: Ok, I have already changed it. :)

marcelina-hasiak (approved):

PR #62, opened 2022-03-05
feat: add MetaData component

wolnio:


app/components/MovieHero/MetaData/index.js
```diff
@@ -0,0 +1,26 @@
...
+import PropTypes from 'prop-types';
+
+export default function MetaData({...props}) {
+//NOTE: real Movie data from an API will be added later, when the upper comonent MovieHero will be edited
+  const movieData = {
+    title: 'Fight Club',
+    release_date: '1999-10-15',
+    rate: 8.4,
+    popularity: 60,
+  };
```
marcelina-hasiak, line 11: You could use `movieMockData.json` file and just import it.

app/components/MovieHero/MetaData/MetaData.styles.js
```diff
@@ -0,0 +1,39 @@
+import styled from 'styled-components';
+import { device } from '../../../consts/mediaQueries';
+
+const MetaDataWrapper = styled.div`
+  color: #fff;
+  font-size: 14px;
+  margin: ${({ margin }) => margin || '0 0 0 0'};
```
marcelina-hasiak, line 7: I would use only margin-bottom here (spacing between title and MetaData you'll achieve by adding margin-bottom to title), but if you like to keep margin-top as well - no problem. However, horizontal margin are unnecessary.

app/components/MovieHero/MetaData/MetaData.styles.js
```diff
@@ -0,0 +1,39 @@
...
+
+const MetaDataItem = styled.span`
+  padding: 0 16px;
+  &:first-child {
+    padding-left: 0;
+  }
+  &:last-child {
+    padding-right: 0;
+  }
+`;
```
marcelina-hasiak, line 22: Simply padding-right is all you need here. Try: 
```
&:not(:last-child) {
  padding-right: 16px
}
```

marcelina-hasiak (changes requested):

app/components/MovieHero/MetaData/MetaData.styles.js
```diff
@@ -0,0 +1,39 @@
...
+  color: #fff;
+  font-size: 14px;
+  margin: ${({ margin }) => margin || '0 0 0 0'};
+
+  ${device.tablet} {
+    font-size: 20px;
+  }
+`;
+
+const MetaDataItem = styled.span`
```
marcelina-hasiak, line 14: Span hasn't got semantic meaning. It's good to use it as separator but use paragraph to wrap important data.

app/components/MovieHero/MetaData/index.js
wlechowicz: This file needs to go away, actually I should have deleted it when I added first 2 movies endpoints.

app/components/MovieHero/MetaData/MetaData.styles.js
wlechowicz: The outer-most component shouldn't have any margins at all. If anything, it should use padding for internal spacing. This will make composing this component with others easier.

app/components/MovieHero/MetaData/index.js
wlechowicz, line 11: Please don't use mock data in the component itself. You can use it in the story to showcase how the component works.

app/components/MovieHero/MetaData/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+  return (
+    <MetaDataWrapper {...props}>
```
wlechowicz, line 14: You're passing entire `props` object to child, this means that most likely something's wrong with composition or logic placement - as a general rule. The wrapper only uses `margin` prop, so only that prop should be passed.

app/components/MovieHero/MetaData/index.js
```diff
@@ -0,0 +1,26 @@
...
+      <MetaDataItem>{movieData.release_date.slice(0, 4)}</MetaDataItem>
+      <MetaDataSeparator />
+      <MetaDataItem>{movieData.rate}/10</MetaDataItem>
+      <MetaDataSeparator />
+      <MetaDataItem>{movieData.popularity}</MetaDataItem>
+    </MetaDataWrapper>
```
wlechowicz, line 20: That's a lot of `MetaData` in the names, I think you should drop this and just name them `Separator`, `Wrapper`, `Item`. They are local to your component so the generic naming is fine.

app/components/MovieHero/MetaData/MetaData.stories.js
```diff
@@ -0,0 +1,13 @@
+import MetaData from './';
```
wlechowicz, line 1: I suggest renaming the whole component to `MovieMetadata`.

app/components/MovieHero/MetaData/MetaData.styles.js
```diff
@@ -0,0 +1,39 @@
+import styled from 'styled-components';
+import { device } from '../../../consts/mediaQueries';
+
+const MetaDataWrapper = styled.div`
+  color: #fff;
+  font-size: 14px;
```
wlechowicz, line 6: Font size should be either relative to its surroundings (`em`), or constant but with `rem`.

app/components/MovieHero/MetaData/MetaData.styles.js
wlechowicz: `p` would be a bit annoying to style and wouldn't be very fitting semantic-wise. The best solution for this whole component would be to use a list `ul` and make it inline.

wlechowicz (changes requested):
Like I said in standup, this component should not be located inside MovieHero, because it will be used in multiple locations. Please move it to components/UI.

app/components/MovieHero/MetaData/MetaData.styles.js
wolnio: Should I stick to the design and do some unit conversion, so `font-size` with `em/rem` would have similar size in `px`?

app/components/MovieHero/MetaData/MetaData.styles.js
wolnio: Well, I need some space on both sides of my 'bullet-like' separator. This code will give me space only on the right side.

app/components/MovieHero/MetaData/MetaData.styles.js
marcelina-hasiak: Exactly. Remember round this values if needed, we're not pixel perfect here ;)

marcelina-hasiak: Exactly. Remember round this values if needed, we're not pixel perfect here ;)

app/components/UI/MovieMetadata/MovieMetadata.stories.js
```diff
@@ -0,0 +1,19 @@
...
+  release_date: '1999-10-15',
+  popularity: 60.21,
+  vote_average: 8.4,
+};
+
+const Template = ({ ...args }) => <MovieMetadata {...args} {...metaDataObj} />;
+
+export const Default = Template.bind({});
+Default.args = {
+  marginBottom: '0',
```
weronika299, line 18: You should use it in MovieMetadata.styles.js

app/components/UI/MovieMetadata/index.js
```diff
@@ -0,0 +1,16 @@
+import { Wrapper, Item, Separator, ItemList } from './MovieMetadata.styles';
+import PropTypes from 'prop-types';
```
weronika299, line 2: You declared it but you didn't use it.

weronika299 (changes requested):

app/components/MovieHero/MovieHero.stories.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
wlechowicz, line 1: Not needed.

app/components/UI/MovieMetadata/MovieMetadata.styles.js
```diff
@@ -0,0 +1,38 @@
+import styled from 'styled-components';
+import { device } from '../../../consts/mediaQueries';
+
+const Wrapper = styled.div`
+  color: #fff;
```
wlechowicz, line 5: `color` should be inherited from parent, this way this component will work anywhere it's used.

app/components/UI/MovieMetadata/MovieMetadata.styles.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+`;
+
+const Separator = styled.span`
+  width: 4px;
+  height: 4px;
+  background-color: #44ec52;
+  border-radius: 50%;
+  display: inline-block;
+  top: 50%;
```
wlechowicz, line 28: This should have no effect since it's inline.

app/components/UI/MovieMetadata/index.js
```diff
@@ -0,0 +1,16 @@
...
+
+export default function MovieMetadata({ ...props }) {
```
wlechowicz, line 4: `({ ...props })` does nothing, it's the same as `(props)`. You should destructure the props that you need with `({ release_date, vote_average, popularity })`

wlechowicz (changes requested):
This PR is for MovieMetadata component only, MovieHero component should not be a part of it.

app/components/UI/MovieMetadata/MovieMetadata.stories.js
wolnio: Thats right, forgot to remove it after latest review

app/components/UI/MovieMetadata/MovieMetadata.styles.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+`;
+
+const Separator = styled.span`
+  width: 4px;
+  height: 4px;
+  background-color: #44ec52;
+  border-radius: 50%;
+  display: inline-block;
+  transform: translateY(-50%);
```
rast-mk1, line 25: Why do you need transform?

weronika299 (approved):
Looks good to me.

app/components/UI/MovieMetadata/index.js
```diff
@@ -0,0 +1,13 @@
+import { Item, Separator, ItemList } from './MovieMetadata.styles';
+
+export default function MovieMetadata({ release_date, vote_average, popularity }) {
```
wlechowicz, line 3: The data should come in as a single prop like `movieData` and then be further destructured into those fields. When destructuring, please change snake_case to camelCase.

wlechowicz (changes requested):

app/components/UI/MovieMetadata/MovieMetadata.styles.js
marcelina-hasiak: it's same as `vertical-align: middle` - center the bullet

marcelina-hasiak (approved):

app/components/UI/MovieMetadata/index.js
wolnio: Ok, I wanted to keep variable names as they are in JSON response from movie API, but will change it :)

wlechowicz: Yeah it really is something that should be done on the backend. @rast-mk1 can you add the snake case to camel case transform to movieDb client? Like this https://stackoverflow.com/a/61375162

wlechowicz (approved):

PR #63, opened 2022-03-07
feat: redesign input components

weronika299:


app/components/Form/Input/Input.styles.js
```diff
@@ -1,12 +1,37 @@
...
+    display: none;
+  }
 `;
 
-export default StyledInput;
+export const ErrorMessage = styled.div`
+  position: absolute;
+  bottom: -12px;
+  line-height: 20px;
+  font-size: 14px;
```
marcelina-hasiak, line 35: It's been already said in different PR to use em or rem units for font-size

marcelina-hasiak (changes requested):

app/components/Form/Input/index.js
```diff
@@ -1,14 +1,25 @@
...
+    <InputWrapper>
+      <StyledInput
+        value={value}
+        onChange={onInputChange}
+        name={name || id}
+        id={id}
+        error={error}
+        {...props}
+      />
+      {label && <InputLabel htmlFor={id}>{label}</InputLabel>}
```
marcelina-hasiak, line 15: As far as I know input always should has a label even if it's hidden. I would remove this conditional rendering - I didn't catch this previously and sorry for that ^^'

app/components/Form/Input/Input.styles.js
```diff
@@ -1,12 +1,52 @@
...
+  border-radius: 8px;
+  background-color: #272233;
+  color: #e9e9e9;
+  &:focus + ${InputLabel} {
+    display: none;
+  }
+  &:not([value='']) + ${InputLabel} {
+    font-size: 0.8rem;
+    transform: translateY(-2rem);
+  }
```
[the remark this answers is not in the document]
weronika299: I need this to show the label when I have text in the input. 
password input
figma input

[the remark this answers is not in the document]
marcelina-hasiak: Ok, I know what you mean but you have something like this:   

```
input:focus + ${InputLabel} {
    display: none;
 }
```
 1. I would remove hiding label on focus input entirely and animate this floating label by adding some transition
 2. Input value and label need more placement adjusting:
Zrzut ekranu 2022-03-9 o 20 22 21
Zrzut ekranu 2022-03-9 o 20 23 41
3. About storybook... Label doesn't float because input state isn't update and the value equals "" all the time. I tried to pass some function inside inInputChange arg but it didn't help. I don't how to solve it and if solution is needed at all ??

app/components/Form/Input/Input.stories.js
```diff
@@ -1,22 +1,34 @@
...
-const Template = (args) => <Input {...args} />;
+const Template = (args) => {
+  const [value, setValue] = useState(args.value ?? '');
+  return (
+    <Input
+      {...args}
+      onChange={(e) => {
+        args.onInputChange(e.target.value);
+        setValue(e.target.value);
+      }}
```
marcelina-hasiak, line 16: Don't need to call args.onInputChange(e.target.value) here in story, onChange={(e) => setValue(e.target.value)} will be enough for our purpose.

app/components/Form/Input/Input.stories.js
```diff
@@ -1,22 +1,34 @@
...
-
+import { useState } from 'react';
 export default {
   title: 'Form/Input',
   component: Input,
 };
 
-const Template = (args) => <Input {...args} />;
+const Template = (args) => {
+  const [value, setValue] = useState(args.value ?? '');
```
marcelina-hasiak, line 9: const [value, setValue] = useState(args.value || '')

marcelina-hasiak (changes requested):
Input component with its own state in story works good now. Don't know other/better solution how to present label changing state. To be continued on daily or with @wlechowicz earlier here.

app/components/Form/Input/Input.stories.js
weronika299: Why not double question mark? I set value on args.value but if it is null or undefined I set ''.

marcelina-hasiak: Hmmmm... I compared it with || and didn't see any differences. Could you explain it to me in DM? I’m surely missing the point.

marcelina-hasiak (approved):

hiimsagan (approved):
Looks Good

wlechowicz (approved):

PR #65, opened 2022-03-11
feat: add method to get images from theMovieDB #1406 

rast-mk1:
**NEW APPROACH**

Possible use:

> // pages/index.js

```
import styles from '../styles/Home.module.css';

export default function Home({ movie }) {
  const poster = movie.images.poster.l;

  return (
    <div className={styles.container}>
      <main className={styles.main}>
        <h1>{movie.title}</h1>
        <p>{movie.overview}</p>
        <img
          style={{ objectFit: 'contain' }}
          src={poster}
          title={movie.title}
          alt={movie.title}
          width={'500'}
          height={'500'}
        />
      </main>
    </div>
  );
}

export async function getServerSideProps() {
  const movieId = 76341;
  const resp = await fetch(`http://localhost:3000/api/movies/${movieId}`);
  const data = await resp.json();
  return { props: { movie: data } };
}

```


wlechowicz:
There was a misunderstanding as to how to approach this, details discussed in slack.

wlechowicz (approved):
Overall very nice iterative improvements, according to guidance. Last minor change to be done: `apiResponse` isn't a good name since response (`res`) is already defined, and after `res.json();` we're left with just the parsed response body data. `apiResponse` is a bit misleading since a response would be expected to have a specific structure, like `.ok` field etc.

PR #66, opened 2022-03-11
fix: allow setting icon custom size inside button

marcelina-hasiak:


app/components/UI/Button/Button.styles.js
```diff
@@ -47,38 +47,28 @@ const StyledButton = styled.button`
...
 
     ${({ iconWithBg }) =>
     iconWithBg &&
     css`
        {
-        background: #1e1b26;
+        background-color: #1e1b26;
         border-radius: 50%;
-        padding: 1rem;
+        padding: ${({ padding }) => padding || '1.4rem'};
```
hiimsagan, line 69: you should keep 1rem on basic becouse you will change all buttons i

hiimsagan (changes requested):

app/components/UI/Button/Button.styles.js
marcelina-hasiak: 1. Not all buttons, but only buttons with icon and round background
2. Yeah, button is as big as should be on desktop. I forgot we use mobile-first approach ^^'

hiimsagan (approved):

PR #67, opened 2022-03-13
fix: form display on mobile devices

weronika299:


marcelina-hasiak (changes requested):
Zrzut ekranu 2022-03-13 o 19 56 05
I know this fix is related to mobile only, but can we fix desktop view as well? After adding form component I met some issues: 
1. Missing padding top and bottom
2. Way to big logo, I tested some values and `width: clamp(70px,10vw,100px)` added for StyledLogoLink component fits good even for header
3. Missing spacing below Sign-up and Sign-in tabs
* I think input appearance where all 3 inputs display error message is too much. Try to set green border 1px width and error messages to font-size: 500 or 400 and check the outcome

wlechowicz:
Definitely the rectangle container should have an all-around padding to give some breathing room for its contents.
Also the logo shouldn't be clickable, there's no point in redirecting to home page from this screen.

wlechowicz (approved):

marcelina-hasiak (approved):

PR #68, opened 2022-03-13
feat: create SearchMovie component

marcelina-hasiak:
Component'll be adapt to its parent after header merging.


wlechowicz (changes requested):
There's something odd happening with the icon. When you click the icon and the search is expanding, the icon seems to disappear briefly, then it comes back. It works fine when it's collapsing, though.

wlechowicz (approved):

PR #70, opened 2022-03-13
feat: user authorization using next-auth

wlechowicz:
Implements next-auth with JWT and users stored in MongoDB on Atlas.
LoginForm used as a POC of a working sing-in. You can sign in to Preview with john@test.com / aaabbb and after you're signed in, clicking on user avatar will sign you out.


app/components/Form/LoginForm/index.js
```diff
@@ -1,43 +1,42 @@
...
-    setUsernameError('');
-    if (!username || username.length < 5) {
-      setUsernameError('Invalid username');
-    }
+    setLoginError(false);
 
-    const user = { username, password };
-    console.log(user);
+    const user = { email, password };
+    const res = await signIn('credentials', { ...user, callbackUrl: '/', redirect: false });
```
wlechowicz, line 20: `signIn` comes from next-auth, by default it would auto-redirect to error page on error, but we want to display the error in the form so we prevent that with `redirect: false`. This means we need to handle `res.error` ourselves.

app/components/Header/index.js
```diff
@@ -1,20 +1,14 @@
...
+import { MainNavigationBar, UserPanel, MobileMenuBtn, UserImg, NavigationBar } from './headerStyle';
 import LogoLink from 'components/UI/LogoLink';
-import Icon from 'components/Icon';
 
 import { useState } from 'react';
+import { signOut, useSession } from 'next-auth/react';
 
 export default function Header() {
   const [isMobileMenuVisible, setMobileMenu] = useState(false);
+  const { data: session } = useSession();
```
wlechowicz, line 11: `useSession` hook tells us if user is logged in or not.

app/components/Header/index.js
```diff
@@ -26,11 +20,8 @@ export default function Header() {
           <MainNav />
         </MainNavigationBar>
         <UserPanel>
-          {/* TODO search input */}
-          <UserImg>{userImg}</UserImg>
-          <ShowAccount>
-            <Icon type="sort-down" />
-          </ShowAccount>
+          {session.user?.name}
+          <UserImg onClick={() => signOut()}>{userImg}</UserImg>
```
wlechowicz, line 24: `signOut` comes from next-auth and does just that - signs the user out.

app/lib/hash.js
```diff
@@ -0,0 +1,14 @@
+import { createHash, randomBytes } from 'crypto';
+
+export function getPassHash(password) {
+  const salt = process.env.PASS_SALT || '';
+  const hash = createHash('sha256', salt).update(password).digest('base64');
+  return hash;
+}
```
wlechowicz, line 7: We're storing just a hash of the password in database, so when user logs in, we hash the password and compare the hashes.

pages/_app.js
```diff
@@ -3,16 +3,19 @@ import '../styles/fonts.css';
...
-  const { pageLayout } = pageProps;
+  const { pageLayout, session } = pageProps;
   const [appState] = useState({});
   return (
-    <AppContext.Provider value={appState}>
-      <Layout pageLayout={pageLayout}>
-        <Component {...pageProps} />
-      </Layout>
-    </AppContext.Provider>
+    <SessionProvider session={session}>
```
wlechowicz, line 12: This provides the session data to the whole app, which we can use with the `useSession` hook.

pages/api/movies/[id].js
```diff
@@ -22,6 +23,10 @@ import { getMovieById } from '../../../lib/services/movieDb';
  *         description: movie not found
  */
 export default async function handler(req, res) {
+  const session = await getSession({ req });
+  if (!session) {
+    return res.status(401).send('Not authorized');
+  }
```
Comment (wlechowicz) [inline pages/api/movies/[id].js:29]: This is how easy it is to make an API route available to logged in users only.

pages/api/users/login.js
```diff
@@ -0,0 +1,27 @@
...
+
+  if (!(email && password)) {
+    return res.status(400).json({ error: 'Missing required fields' });
+  }
+
+  const user = (await db.collection('users').find({ email }, { limit: 1 }).toArray())[0];
+
+  const passHash = getPassHash(password);
+
+  if (!user || user.passHash !== passHash) {
```
wlechowicz, line 22: Comparing the password hashes. We don't store the password in DB so it's impossible to steal it.

pages/api/users/register.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+  const passHash = getPassHash(password);
+
+  const newUser = {
+    username,
+    email,
+    passHash,
+    id: randomUUID(),
+    createdAt: new Date(),
+  };
```
wlechowicz, line 32: Here we create the password hash and store that in DB, we never store the actual password.

pages/index.js
```diff
@@ -1,23 +1,19 @@
...
-          isModalOpen: {`${state.isModalOpen}`} <br />
-          <button onClick={() => openModal('Hi')}>action openModal</button>{' '}
-          <button onClick={closeModal}>action closeModal</button>
-        </p>
-      </main>
-    </div>
+    <main
+      style={{ height: '100vh', display: 'flex', alignItems: 'center', justifyContent: 'center' }}
+    >
+      {session?.user ? (
```
wlechowicz, line 10: `session.user` contains username, email and image - very limited set of user data that is safe and cannot be used with malicious intent.

hiimsagan (approved):

weronika299 (approved):

rast-mk1 (approved):

PR #72, opened 2022-03-14
feat: sign-in component

weronika299:
Mobile view is fixed in a separate PR that fixed AuthenticationPageContainer.


rast-mk1 (commented):
What do you think about button having some kind of change in styling while it is being clicked? Not sure if this should be done in this PR, though.

app/components/Form/SignInForm/index.js
```diff
@@ -0,0 +1,66 @@
+import { useState } from 'react';
+import Button from 'components/UI/Button';
+import Input from '../Input';
+import { FormContainer } from './SignInForm.styles';
+
+const emailPattern = /^[a-z\d]+[\w\d.-]*(\+[\w\d.-]*)?@(?:[a-z\d]+[a-z\d-]+\.){1,5}[a-z]{2,6}$/i;
```
wlechowicz, line 6: This is too strict. Something more generic like "looks like an email more or less, so has at least 5 characters and contains @ and ."
We need to be that strict when registering the user, but when logging in we just want to be a little safe.

wlechowicz (changes requested):

app/components/Form/SignInForm/SignInForm.styles.js
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components';
+
+export const FormContainer = styled.div`
+  min-width: 200px;
+  width: 400px;
+  max-width: 100%;
+  padding-top: 40px;
+`;
```
marcelina-hasiak, line 8: I need it as well. I put this component inside Form folder in form.styles.js file for now.

app/components/Form/SignInForm/index.js
```diff
@@ -0,0 +1,66 @@
...
+      setEmailError('Invalid email format.');
+    }
+
+    if (!password) {
+      isErrorOccured = true;
+      setPasswordError('Please enter your password.');
+    } else if (password.length < 5) {
+      isErrorOccured = true;
+      setPasswordError('Enter at least 5 characters.');
+    }
```
marcelina-hasiak, line 34: We have already validateFormInputs.js file below utils folder. It would be neat to move your validation there.

marcelina-hasiak (changes requested):

app/components/Form/SignInForm/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+    if (!isErrorOccured) {
```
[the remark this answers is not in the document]
wlechowicz: Well no it can't, state updates are queued so we need to treat them as async, we can't assume those states changes synchronously like this. I suggested a different approach instead, with errors being collected in an array locally, before being set in state.

app/components/Form/SignInForm/SignInForm.styles.js
weronika299: Ok, I did it also. We can do this in this way that the person who will merge as the second will delete it.

app/components/Form/SignInForm/index.js
wlechowicz: This validation should be much more relaxed than the sign up one, as explained on stand up.

app/components/Form/SignInForm/index.js
```diff
@@ -1,40 +1,23 @@
...
-const emailPattern = /^[a-z\d]+[\w\d.-]*(\+[\w\d.-]*)?@(?:[a-z\d]+[a-z\d-]+\.){1,5}[a-z]{2,6}$/i;
+const initialState = { email: '', password: '' };
 
 export default function SignInForm() {
-  const [email, setEmail] = useState('');
-  const [password, setPassword] = useState('');
-  const [emailError, setEmailError] = useState('');
-  const [passwordError, setPasswordError] = useState('');
+  const [inputValues, setInputValues] = useState(initialState);
+  const [inputErrors, setInputErrors] = useState(initialState);
```
wlechowicz, line 11: Not a fan of this initial state for errors as explained in the sign up PR.

app/components/Form/SignInForm/index.js
```diff
@@ -1,40 +1,23 @@
...
-      isErrorOccured = true;
-      setPasswordError('Please enter your password.');
-    } else if (password.length < 5) {
-      isErrorOccured = true;
-      setPasswordError('Enter at least 5 characters.');
-    }
-
-    if (!isErrorOccured) {
-      const user = { email, password };
+    setInputErrors({ initialState });
```
wlechowicz, line 15: Too many curly braces, initialState is an object already.

app/utils/validateFormInputs.js
```diff
@@ -35,3 +35,18 @@ export const validateUserPassword = (password) => {
...
+  const validationErrors = {};
+  if (!email) {
+    validationErrors.email = 'Please enter your email.';
+  } else if (!emailPatternSignIn.test(email)) {
+    validationErrors.email = 'Invalid email format.';
+  }
+  if (!password) {
+    validationErrors.password = 'Please enter your password.';
+  } else if (password.length < 5) {
+    validationErrors.password = 'Enter at least 5 characters.';
```
wlechowicz, line 49: 'Password must contain at least 5 characters'.

wlechowicz (changes requested):

app/components/Form/SignInForm/index.js
```diff
@@ -0,0 +1,51 @@
...
+const initialState = { email: '', password: '' };
+
+export default function SignInForm() {
+  const [inputValues, setInputValues] = useState(initialState);
+  const [inputErrors, setInputErrors] = useState(initialState);
+
+  const userLogin = async (e) => {
+    e.preventDefault();
+    setInputErrors({ initialState });
+    let validationErrors = validateSignInForm(inputValues.email, inputValues.password);
```
marcelina-hasiak, line 16: I would only adapt validateSignInForm to pass only obj inputValues and not its properties but it's just a suggestion.

marcelina-hasiak (approved):
With fixes on Wojtek's comments everything looks good ;)

wlechowicz (approved):

PR #74, opened 2022-03-14
feat: redesign SignUp component

marcelina-hasiak:


app/components/Form/UserSignUpForm/index.js
```diff
@@ -0,0 +1,103 @@
...
+    const isFormValid = !validateInputs();
+    if (isFormValid) {
+      handleFormData();
+    }
+  };
+
+  const validateInputs = () => {
+    const errorMessages = validateSignUpFormInputs(inputValues);
+    const isError = Object.values(errorMessages).some((input) => input !== '');
+    isError ? setErrorMsg(errorMessages) : setErrorMsg(initialState);
```
wlechowicz, line 44: It would be slightly easier and more elegant if you initialized the errors as empty object, then `errorMsg[name]` would be `undefined` or a string. Your condition would then change to `.some((error) => Boolean(error));`.
Also that ternary should be inside like `setErrorMsg(isError ? errorMessages : {});`

app/components/Form/UserSignUpForm/index.js
```diff
@@ -0,0 +1,103 @@
...
+} from 'utils/validateFormInputs';
+import { FormContainer } from '../Form.styles';
+import Input from '../Input';
+import Button from 'components/UI/Button';
+
+const initialState = { email: '', password: '', name: '' };
+
+const UserSignUpForm = () => {
+  const [inputValues, setInputValues] = useState(initialState);
+  const [errorMsg, setErrorMsg] = useState(initialState);
```
wlechowicz, line 15: The name `errorMsg` could be better.
1. it's multiple errors not one,
2. it's not just messages, it represents the errors themselves
I'd suggest just `inputErrors`.

app/components/Form/UserSignUpForm/index.js
```diff
@@ -0,0 +1,103 @@
...
+import validateSignUpFormInputs, {
+  validateUserEmail,
+  validateUserPassword,
+  validateUserName,
+} from 'utils/validateFormInputs';
+import { FormContainer } from '../Form.styles';
+import Input from '../Input';
+import Button from 'components/UI/Button';
+
+const initialState = { email: '', password: '', name: '' };
```
wlechowicz, line 11: Since this should be immutable by the very definition, you can freeze it with https://developer.mozilla.org/pl/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

app/components/Form/UserSignUpForm/index.js
```diff
@@ -0,0 +1,103 @@
...
+  const handleBlur = (event, errorHandler) => {
+    const { name } = event.target;
+    const error = errorHandler(inputValues[name]);
+    setErrorMsg({ ...errorMsg, [name]: error });
+  };
+
+  const handleSubmit = (event) => {
+    event.preventDefault();
+
+    const isFormValid = !validateInputs();
```
wlechowicz, line 35: Usually the convention is for the validate function to return true if valid, false otherwise. I'd suggest flipping the bool in the return statement of `validateInputs` rather than here.

wlechowicz (changes requested):
Nice overall, but some code style changes requested.

weronika299 (approved):
It looks good ;) You can only change the variable names in validateFormInputs because we will have the validation for signIn there as well, for example: passwordPattern to passwordSignUpPattern because signIn has a different regExp for password but that's just a suggestion.

wlechowicz (approved):

app/components/Form/UserSignUpForm/index.js
```diff
@@ -0,0 +1,104 @@
...
+  validateSignUpUserEmail,
+  validateSignUpUserPassword,
+  validateSignUpUserName,
+} from 'utils/validateFormInputs';
+import { FormContainer } from '../Form.styles';
+import Input from '../Input';
+import { SignUpButton } from './UserSignUpForm.styles';
+
+const initialState = { email: '', password: '', name: '' };
+Object.freeze(initialState);
```
wlechowicz, line 12:
```suggestion
const initialState = Object.freeze({ email: '', password: '', name: '' });
```

wlechowicz (changes requested):
Spotted one mistake.

app/components/Form/UserSignUpForm/index.js
wlechowicz: @marcelina-hasiak

PR #75, opened 2022-03-15
feat: backend movies/category/[id] endpoint #1737

rast-mk1:
** **IMPORTANT** ** **this PR will change snake_case notation to camelCase for JSON returned from theMovieDB (first level of object)**

Test this endpoint using /api-doc. 

Outcome of possible use:
obraz


pages/api/movies/category/[id].js
```diff
@@ -0,0 +1,36 @@
...
+ *       404:
+ *         description: movie not found
+ */
+export default async function handler(req, res) {
+  try {
+    const { id } = req.query;
+    const movie = await getListOfMoviesByCategoryId(id);
+    if (!movie) {
+      return res.status(404).json();
+    }
```
Comment (wlechowicz) [inline pages/api/movies/category/[id].js:30]: Since it's not a movie but a list, this const could have a more fitting name.

lib/services/movieDb.js
```diff
@@ -60,7 +63,10 @@ const movieDbClient = async (url, method = 'GET', body) => {
     if (data.backdrop_path) {
       data.images.backdrop = getFullImageURLs(data.backdrop_path);
     }
+    const dataToCamelCase = Object.keys(data).map((key) => [toCamelCase(key), data[key]]);
+    data = Object.fromEntries(dataToCamelCase);
```
wlechowicz, line 67: Since this code is duplicated here and in lines 54-55,, it should become a function that takes in the data to transform and returns a transformed object, like `return makeKeysCamelCase(movie);`

lib/services/movieDb.js
```diff
@@ -12,15 +12,16 @@ class HTTPResponseError extends Error {
...
  * @param {Object} [body] request body
  * @return {Promise<Object|null>} response body
  * @throws {HTTPResponseError}
  */
-const movieDbClient = async (url, method = 'GET', body) => {
+const movieDbClient = async (url, query, method = 'GET', body) => {
   const token = process.env.MOVIEDB_API_KEY;
   const baseUrl = 'https://api.themoviedb.org/3';
-  const wholeUrl = `${baseUrl}/${url}?api_key=${token}&language=en-US`;
+  const wholeUrl = `${baseUrl}/${url}?api_key=${token}&language=en-US${query}`;
```
wlechowicz, line 24: If parameter `query` is not passed in (it's undefined), this will produce a string `&language=en-USundefined`. Default `query` parameter to an empty string in the function declaration.

wlechowicz (changes requested):

pages/api/movies/category/[id].js
Comment (rast-mk1) [reply pages/api/movies/category/[id].js:30]: Agreed

lib/services/movieDb.js
rast-mk1: Good call. Made changes.

lib/services/movieDb.js
rast-mk1: Done as requested.

lib/services/movieDb.js
```diff
@@ -49,8 +50,9 @@ const movieDbClient = async (url, method = 'GET', body) => {
         movie.images.poster = getFullImageURLs(movie.poster_path);
       }
       if (movie.backdrop_path) {
-        movie.images.backdrop = getFullImageURLs(data.poster_path);
+        movie.images.backdrop = getFullImageURLs(movie.backdrop_path);
       }
+      movie = makeKeysCamelCase(movie);
       return movie;
```
wlechowicz, line 56: No need to assign to a variable, you can just `return makeKeysCamelCase(movie);`

lib/services/movieDb.js
```diff
@@ -61,7 +63,9 @@ const movieDbClient = async (url, method = 'GET', body) => {
     if (data.backdrop_path) {
       data.images.backdrop = getFullImageURLs(data.backdrop_path);
     }
+    data = makeKeysCamelCase(data);
```
wlechowicz, line 66: The `data` can stay a const and you can just `return makeKeysCamelCase(data);` here.

lib/services/movieDb.js
```diff
@@ -79,6 +83,16 @@ export function getFullImageURLs(imgUrl) {
...
 }
 
+export function makeKeysCamelCase(data) {
+  const toCamelCase = (str) =>
+    str
+      .toLowerCase()
+      .replace(/([-_][a-z])/g, (group) => group.toUpperCase().replace('-', '').replace('_', ''));
+  const dataToCamelCase = Object.keys(data).map((key) => [toCamelCase(key), data[key]]);
+  data = Object.fromEntries(dataToCamelCase);
+  return data;
```
wlechowicz, line 93: Again, you can just `return Object.fromEntries(dataToCamelCase);`
Assigning to function parameters is not a good idea anyways.

lib/services/movieDb.js
```diff
@@ -13,15 +13,16 @@ class HTTPResponseError extends Error {
  * The Movie Database API Client
  *
  * @param {string} url the TMDB endpoint you want to talk to
+ * @param {string|number} query additional query
```
wlechowicz, line 16: When is `query` a number?

wlechowicz (changes requested):

lib/services/movieDb.js
rast-mk1: At first, changes in movieDbClient were supposed to look differently and having query as number was an option. Forgot to change this after going for different approach.

lib/services/movieDb.js
rast-mk1: True, made changes.

lib/services/movieDb.js
rast-mk1: Yes, that's a good idea.

lib/services/movieDb.js
rast-mk1: Sure. Made changes here.

weronika299 (approved):
I think that it looks good.

wlechowicz (approved):

PR #77, opened 2022-03-16
feat: add user avatar

wlechowicz:
Adds UserAvatar component that grabs user avatar from user session. Should only be used for a signed in user, but will display a blank avatar if you use it wrong.

api/users/signup assigns a random avatar on user signup which is then persisted in DB. 10 avatar images are added to the app for now.


hiimsagan (approved):
Beautiful avatars :)

PR #78, opened 2022-03-16
feat: backend genre/movie/list endpoint #1761

rast-mk1:


rast-mk1:
Due to `getFullImageURLs`  inside `movieDbClient` list of genres has now also `images` key. No harm done, I guess, but still, should I do something about it @wlechowicz ?
obraz

wlechowicz (approved):
Seems safe to ignore that empty `images` obj for now. Solution would be to somehow skip the logic early rather than filtering out the empty obj at the end.

weronika299 (approved):

PR #81, opened 2022-03-17
feat: cross-platform auth and route protection

wlechowicz:
API routes can be easily protected with `withAuth`.
api/users/signin now returns the JWT as `token` in the response, which can be used in authorization header or request cookie in requests to other endpoints.

There is a big `chore: cleanup` commit included which removes unnecessary files and fixes some paths. To review the auth stuff, click on either of 2 other commits to not be distracted by the cleanup one.

The cleanup might give you some conflicts on your feature branches but it's a necessary step. Oh well.


rast-mk1 (approved):
I like it a lot

wlechowicz:
You can still review this post-factum, but I need to merge this so that QA is able to test our APIs.

PR #83, opened 2022-03-18
feat: backend movies/search #1743 

rast-mk1:
```
export async function getServerSideProps() {
  const query = 'escape from';
  const resp = await fetch(`http://localhost:3000/api/movies/search?query=${query}`);
  const data = await resp.json();
  return { props: { movie: data } };
```
and get something like:
obraz


lib/services/movieDb.js
```diff
@@ -125,3 +137,27 @@ export async function getListOfSimilarMoviesById(id) {
...
+
+/**
+ * Get a list of movies belonging to given genre
+ *
+ * @export
+ * @param {string|number} id Comma separated value of genre ids that you want to include in the results
+ * @return {Promise<Object|null>}  movie object
+ */
+export async function getListOfMoviesByCategoryId(id) {
+  const query = `&with_genres=${id}`;
```
wlechowicz, line 149: It would be safe to add something like `if (!id) return null;`

lib/services/movieDb.js
```diff
@@ -125,3 +137,27 @@ export async function getListOfSimilarMoviesById(id) {
...
+
+/**
+ * Search for movies.
+ *
+ * @export
+ * @param {string} query Pass a text query to search.
+ * @return {Promise<Object|null>}  movie object
+ */
+export async function getMovieSearchOutcome(queryText) {
+  const query = `&query=${queryText}`;
```
wlechowicz, line 161: Same as above, return null if param is bogus.

wlechowicz (approved):
Nice, 2 minor improvements suggested.

wlechowicz:
LGTM

PR #85, opened 2022-03-20
docs: api docs as a full fledged openapi yaml

wlechowicz:
Gets rid of the pseudo-docs in comments and replaces them with a fully fledged OpenAPI YAML.

The comment-style docs were hard to write and maintain since intellisense, syntax check and auto formatting were impossible, so were using refs. Now all is possible.

VS Code extension "OpenAPI (Swagger) Editor" highly recommended, for intellisense and formatting

/api-doc page works same as before, but better. Now there's an "Authorize" option that adds the Bearer authorization to requests, you don't need to be logged in on the page anymore. Requests are grouped by tags (movies, users) and order in which they show up can be changed. All of that was impossible in the previous tool.

Note: there's a bug in SwaggerUI that prevents JSON payloads from working, which breaks users endpoint when you use the "Try It Out" option. Will see about a fix - it didn't work before, this PR didn't break it.


hiimsagan (approved):

marcelina-hasiak (approved):

weronika299 (approved):

docs/apiDocs.yml
```diff
@@ -0,0 +1,253 @@
...
+      type: array
+      items:
+        $ref: "#/components/schemas/MovieShort"
+    MovieShort:
+      type: object
+      description: "Shortened Movie object returned as part of lists"
+    MovieFull:
+      type: object
+      description: "Full version of the movie object"
+
```
rast-mk1, line 245: What are these for?

wlechowicz: These are component schemas that are referenced elsewhere through `$ref` so you define those in one spot and reuse them throughout the doc.
Normally they would contain the whole schema of those objects but I didn't feel like writing up the whole movie object schema from TMDB, especially since it has more objects nested inside.

rast-mk1: Got it

rast-mk1 (approved):

PR #86, opened 2022-03-20
Feat: display movie popularity as 5 star rating #1736

wolnio:


app/components/UI/MovieMetadata/StarRating.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+const StarSolid = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: ${(props) => (props.vote / 10) * 100}%;
+  white-space: nowrap;
+  overflow: hidden;
+`;
```
rast-mk1, line 29: Clever idea with star filling mechanism! Nonetheless, there is no visual difference between all the scores from 1.6 to 2.4.  I would reccomend either rounding voteAverage to full numbers or reworking existing solution, so the difference between e.g. 1.7 and 1.8 is visible.
obraz
obraz

rast-mk1: btw there is no glitch - just made stars bigger locally

app/components/UI/MovieMetadata/StarRating.js
```diff
@@ -0,0 +1,36 @@
...
+
+const Wrapper = styled.div`
+  color: yellow;
```
wlechowicz, line 32: Yellow is too harsh, please change to `gold`.

wlechowicz (changes requested):
It's a clever solution to span the full stars dynamically like that, but I'm not sure if it's the best one. If you set the vote to something like 5.5, it looks like a glitch, as the star is missing a pixel.
Usually the ratings are displayed as empty, full and half stars to avoid these issues, and we should do the same because a difference of 0.1 is not visually represented. There is a `fa-star-half-stroke` that you can use for the third state.

app/components/UI/MovieMetadata/StarRating.js
wolnio: Wait, why stars look bigger than rest of the text? When I tested it in StoryBook it looked the same size as text. Did you messed around with font-size?

rast-mk1: Yup, I did mess with these at my side and these print screens are from my localhost storybook.

rast-mk1 (changes requested):
You get your first half star when score equals 0.6, the you get whole star when score equals 1.5, then another half with 2.6, then with 3.5, then with 4.6 and so on. This is not consistent.

wolnio:
You have to notice that the average vote from API comes in a scale of 10. The star rating is in a scale of 5 so inside `StarRating.js` I divide voteAverage by 2.

rast-mk1:
> You have to notice that the average vote from API comes in a scale of 10. The star rating is in a scale of 5 so inside `StarRating.js` I divide voteAverage by 2.

Yes, I get that. Each point on a scale 0-10 is represented by half of a star and there are 5 stars. What I have is mind is that numbers from 1.0 to 1.4 should be rounded down to 1 and numbers from 1.5 to 1.9 should be rounded up to 2. It doesn't work like that now.

app/components/UI/MovieMetadata/StarRating.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+    return content;
+  };
+
+  return (
+    <Wrapper>
+      <div>{printStars(5)}</div>
+    </Wrapper>
+  );
+}
```
wlechowicz, line 32: The logic seems very complicated for the purpose and has too many steps. 3 steps should be enough:
- generate full stars
- generate one half star if necessary
- fill the rest with empty stars

Proposed solution below. Function moved outside of component - this way it's constant and doesn't get redeclared on every render, which is more performant.

```suggestion
const printStars = (vote) => {
  // vote is 0 - 10
  // cap score to 10 and divide by 2 to adjust to 5 star rating
  let amount = Math.min(vote, 10) / 2;
  const stars = [];
  // generate full stars
  while (amount >= 0.7) {
    stars.push(<Icon type="star" />);
    amount--;
  }
  // add a half star if necessary
  if (amount >= 0.3) {
    stars.push(<Icon type="starHalf" />);
  }
  // fill the rest (5 - stars.length) with empty stars
  const emptyStars = Array.from({ length: 5 - stars.length }, () => <Icon type="starOutline" />);
  // concat stars and empty stars
  return [...stars, ...emptyStars];
};

export default function StarRating({ vote }) {
  return (
    <Wrapper>
      <div>{printStars(vote)}</div>
    </Wrapper>
  );
}
```

This logic could also be done with one pass over the array with a single Array.reduce which _reduces_ (ba dum tss) the code further, but gets a little hardcore. The suggested code above is a better balance between code quality, performance and readability.
```
const printStars = (vote) => {
  // vote is 0 - 10
  // cap score to 10 and divide by 2 to adjust to 5 star rating
  let amount = Math.min(vote, 10) / 2;

  return Array.from({ length: 5 }).reduce((stars) => {
    let starToAdd = <Icon type="starOutline" />;
    if (amount >= 0.7) {
      starToAdd = <Icon type="star" />;
    } else if (amount >= 0.3) {
      starToAdd = <Icon type="starHalf" />;
    }
    amount--;
    return [...stars, starToAdd];
  }, []);
}
```

wlechowicz (changes requested):

app/components/UI/MovieMetadata/StarRating.js
wlechowicz: As for the ranges of stars vs half-stars, we have 0.3 - 0.4 - 0.3 for empty - half - full which I think is fine.

wlechowicz: Or is it actually 0.3 - 0.4 - 0.6, because the third star will be full in the range of [2.7, 3.3)

wlechowicz: Yeah these thresholds of 0.3 and 0.7 don't work, I've edited my suggestion.

rast-mk1:
For some reason, there are only 4 checks available for your newest commit and Storybook didn't generate overview for us to investigate. Weird.

wlechowicz (approved):

rast-mk1 (approved):

weronika299 (approved):

marcelina-hasiak (approved):
Looks good. 
One thing I noticed (but not related to this PR) and it would be nice to change someday is popularity score, because now user don't know what this number means.

PR #88, opened 2022-03-21
feat: add Signup page logic #619 

marcelina-hasiak:


app/components/Form/UserSignUpForm/index.js
```diff
@@ -33,10 +41,30 @@ const UserSignUpForm = () => {
...
+      signUp();
+    }
+  };
+
+  const signUp = async () => {
+    try {
+      const response = await fetchWrapper.post(signUpURL, inputValues);
+      handleServerResponse(response);
+    } catch (error) {
+      setRegisterError('Something went wrong');
```
wlechowicz, line 57: 'Oops! Something went wrong. Please try again later.'

app/components/Form/UserSignUpForm/index.js
```diff
@@ -88,6 +116,17 @@ const UserSignUpForm = () => {
         onInputChange={handleInputChange}
         onBlur={(e) => handleBlur(e, validateSignUpUserPassword)}
       />
+      {registerError && !isFormSubmitting && (
+        <ServersideMessage error>{registerError}</ServersideMessage>
+      )}
+      {registerError === '' && !isFormSubmitting && (
+        <ServersideMessage>
+          {'Now you can '}
```
wlechowicz, line 124: 'Success! Now you can _sign in_!'

app/components/Form/UserSignUpForm/index.js
```diff
@@ -46,18 +74,18 @@ const UserSignUpForm = () => {
...
+      const timer = setTimeout(() => {
+        if (registerError !== null) {
+          setIsFormSubmitting(false);
+        }
+      }, 1500);
+      return () => {
+        clearTimeout(timer);
+      };
+    }
+  }, [isFormSubmitting, registerError]);
```
wlechowicz, line 88: This doesn't seem necessary anymore.

app/components/Form/UserSignUpForm/index.js
```diff
@@ -33,10 +41,30 @@ const UserSignUpForm = () => {
...
+    } catch (error) {
+      setRegisterError('Something went wrong');
+    }
+  };
+
+  const handleServerResponse = (response) => {
+    const { status } = response;
+    if (status === 404) router.push('/404');
+    if (status === 400 || status === 409) {
+      response.json().then((data) => setRegisterError(data.error));
```
wlechowicz, line 65: This isn't necessarily a good idea. You should handle 400 and 409 differently and with custom error messages.

Cases for 400 are
- missing required fields (payload error)
- invalid values or those fields (validation error)

both of which are an oopsie on your side, not a user fault, so user should see "Oppsie! Something went wrong but it shouldn't have and it's totally not your fault but ours and we've dispatched our most skilled team of monkeys to bang on it with a hammer" type of message.

409 is plain and simple but "Email already in use" is kinda rough. Something like "Email address already in use. Try _signing in_ instead!".
You know, smile and wave, just smile and wave.

app/components/Form/UserSignUpForm/UserSignUpForm.styles.js
```diff
@@ -4,3 +4,15 @@ import Button from 'components/UI/Button';
 export const SignUpButton = styled(Button)`
   margin-top: 40px;
 `;
+
+export const ServersideMessage = styled.p`
+  color: ${({ error }) => (error ? '#ff5c00' : '#44ec52')};
+  position: absolute;
+  font-size: 0.9rem;
+`;
```
wlechowicz, line 12: The style of these messages should be uniform across sign in and sign up. Yes, that means a common component under Form.

app/components/Form/UserSignUpForm/index.js
```diff
@@ -33,10 +41,30 @@ const UserSignUpForm = () => {
...
+      const response = await fetchWrapper.post(signUpURL, inputValues);
+      handleServerResponse(response);
+    } catch (error) {
+      setRegisterError('Something went wrong');
+    }
+  };
+
+  const handleServerResponse = (response) => {
+    const { status } = response;
+    if (status === 404) router.push('/404');
```
wlechowicz, line 63: 404 after signup? If it ever happens - which it shouldn't - then it should not take the user to 404 page, because "I just clicked a button, why does it say not found?" but display an oopsie error.

app/components/Form/UserSignUpForm/index.js
```diff
@@ -88,6 +116,17 @@ const UserSignUpForm = () => {
         onInputChange={handleInputChange}
         onBlur={(e) => handleBlur(e, validateSignUpUserPassword)}
       />
+      {registerError && !isFormSubmitting && (
```
wlechowicz, line 119: `&& !isFormSubmitting` is duplicated, which means it should be like
`{!isFormSubmitting && { registerError ? something : something else }}`

app/components/Pages/404/index.js
```diff
@@ -0,0 +1,22 @@
+import { Wrapper404, Message404, ImageWrapper } from './404.styles';
+import Image from 'next/image';
+import appLogo404Img from '/public/images/app-logo-404.svg';
+
+export default function Layout404() {
+  return (
+    <Wrapper404>
+      <ImageWrapper>
+        <Image
+          src={appLogo404Img}
```
wlechowicz, line 10: Nice touch, might be more fitting for the 500 cases, but I like the extra thought :)

app/utils/fetchWrapper.js
```diff
@@ -0,0 +1,13 @@
...
+};
+
+async function post(url, body) {
+  const requestOptions = {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify(body),
+  };
+  const response = await fetch(url, requestOptions);
+  return response;
```
wlechowicz, line 12: No point in awaiting just to return it, since it would have to be awaited again anyway. Also nice wrapper, but fetch might need to be wrapped in try..catch. And it could also do the res.json() while we're at it, and then `return { ok: res.ok, json };`. This would turn it from "post JSON get Response" into "post JSON get JSON".

wlechowicz (changes requested):
One thing I can't comment inline, is that the field in form should say "address" rather than "adress".

rast-mk1:
In general, I like it.
Something not related to this specific PR, but Form itself: when there is something typed in and you hit refresh, labels and values are mixed together.

wlechowicz:
> In general, I like it. Something not related to this specific PR, but Form itself: when there is something typed in and you hit refresh, labels and values are mixed together.

Also happens with auto-fill. Let's refine this once QA reports it.

app/components/Form/UserSignUpForm/index.js
marcelina-hasiak: Do you mean timeout? But what about case where the response is too quick? I need to extend animation to prevent glitch.

weronika299 (commented):
I have only one suggestion. What do you think about changing style of ServersideMessage: 'Now you can sign in'? I think the 'sign in' link could have a slightly different color than the rest of the message.

app/components/Form/UserSignUpForm/index.js
wlechowicz: Ah I see, forgot about that. Boomer brain.

wlechowicz:
> I have only one suggestion. What do you think about changing style of ServersideMessage: 'Now you can sign in'? I think the 'sign in' link could have a slightly different color than the rest of the message.

Definitely links need to stand out from the rest of the text in all cases, they can be underlined (default a style).

hiimsagan (changes requested):
I thnik you forgot to show error massage

obraz_2022-03-24_230756

marcelina-hasiak:
> I have only one suggestion. What do you think about changing style of ServersideMessage: 'Now you can sign in'? I think the 'sign in' link could have a slightly different color than the rest of the message. 
It's already underlined, bold and uppercase. Do we need more?

> obraz_2022-03-24_230756
I know, I know. TY

app/components/Form/UserSignUpForm/index.js
```diff
@@ -1,22 +1,29 @@
...
 import Input from '../Input';
-import { SignUpButton } from './UserSignUpForm.styles';
+import { SignUpButton, StyledLink } from './UserSignUpForm.styles';
+import { ServersideMessage } from '../Form.styles';
+import { fetchWrapper } from 'utils/fetchWrapper';
+import Icon from 'components/UI/Icon';
 
-const initialState = { email: '', password: '', name: '' };
+const initialState = { email: '', password: '', username: '' };
 Object.freeze(initialState);
```
wlechowicz, line 17:
```suggestion
const initialState = Object.freeze({ email: '', password: '', username: '' });
```

app/components/Form/UserSignUpForm/index.js
```diff
@@ -33,31 +40,57 @@ const UserSignUpForm = () => {
...
-    const timer = setTimeout(() => setIsFormSubmitting(false), 3000);
-    return () => {
-      clearTimeout(timer);
-    };
-  }, [isFormSubmitting]);
+    if (isFormSubmitting) {
+      const timer = setTimeout(() => {
+        if (registerError !== null) {
+          setIsFormSubmitting(false);
+          setInputValues(initialState);
```
wlechowicz, line 86: As I mentioned many times, clearing form on submit error is super annoying to the user, because they must retype everything even if just one field was bad.

app/components/Form/UserSignUpForm/index.js
```diff
@@ -88,6 +121,22 @@ const UserSignUpForm = () => {
         onInputChange={handleInputChange}
         onBlur={(e) => handleBlur(e, validateSignUpUserPassword)}
       />
+      {!isFormSubmitting &&
+        registerError !== null &&
+        (registerError ? (
```
wlechowicz, line 126: That's a bit confusing, `registerError` null means there was no error or nothing was done, any other value means there was an error except for an empty string that means success? It would be easier to have an explicit success flag rather than this combination of `not submitting combined with an error but it actually means success`.

app/components/Form/UserSignUpForm/index.js
```diff
@@ -88,6 +121,22 @@ const UserSignUpForm = () => {
...
+            {registerError}
+          </ServersideMessage>
+        ) : (
+          <ServersideMessage>
+            <Icon type="check-mark" style={{ padding: '0 5px' }} />
+            {'Success! Now you can '}
+            <Link href="/sign-in" passHref>
+              <StyledLink>sign in</StyledLink>
+            </Link>
+          </ServersideMessage>
```
wlechowicz, line 138: I suggested a different solution with an outline instead. You can extrapolate that with a green outline and another prop that makes shows the outline and success message.

app/utils/fetchWrapper.js
```diff
@@ -0,0 +1,17 @@
...
+
+async function post(endpoint, body) {
+  const requestOptions = {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify(body),
+  };
+  try {
+    const response = await fetch(`${process.env.NEXT_PUBLIC_API_URL}/${endpoint}`, requestOptions);
+    return { status: response.status, data: response.json() };
```
wlechowicz, line 13: Hmm does this work? The function `post` will return a Promise that resolves to this object with `status` and `data`, but `data` will still be a Promise of the `json()` method result.
You should await the JSON to be parsed like 
```
 const data = await response.json();
 return { status: response.status, data };
```

wlechowicz (changes requested):

app/components/Form/UserSignUpForm/index.js
marcelina-hasiak: I know, I added this code recently, because this is how it works on merged sign-in - I thought this is the way.

app/utils/fetchWrapper.js
marcelina-hasiak: Yep, my mistake. I don't use data inside sign-up so I didn't catch it.

app/components/Form/UserSignUpForm/index.js
marcelina-hasiak: I don't like it as well - it's not readable, but I didn't know how to merge in one all these logical statements.

app/components/Form/ServerSideMessage/index.js
```diff
@@ -0,0 +1,17 @@
+import PropTypes from 'prop-types';
+import { StyledServerSideMessage, ServerSideMessageIcon } from './ServerSideMessage.styles';
+
+const ServerSideMessage = ({ children, errorType, ...props }) => {
+  return (
+    <StyledServerSideMessage errorType={errorType} {...props}>
```
wlechowicz, line 6: `errorType` is misleading, seems to suggest that the prop contains information about the type of error, while in fact it's a boolean and should be called `hasError` or `isError`, the former being the one used elsewhere so I'd stick to that.

app/components/Form/ServerSideMessage/index.js
```diff
@@ -0,0 +1,17 @@
...
+      <ServerSideMessageIcon type={errorType ? 'x-mark' : 'check-mark'} />
```
wlechowicz, line 7: An improvement to this would be to encapsulate this icon type logic in the component itself, like this
```
export const ServerSideMessageIcon = styled(Icon).attrs(
  ({ hasError }) => ({ type: hasError ? 'x-mark' : 'check-mark' })
)`
  padding: 0 5px;
`;
```

app/components/Form/UserSignUpForm/index.js
```diff
@@ -1,22 +1,34 @@
...
+const on409ErrorMessage = 'Email address already in use. Try signing in instead.';
 
 const UserSignUpForm = () => {
   const [inputValues, setInputValues] = useState(initialState);
   const [inputErrors, setInputErrors] = useState({});
   const [isFormSubmitting, setIsFormSubmitting] = useState(false);
+  const [registerError, setRegisterError] = useState(null);
+
+  const onSuccess = Boolean(!isFormSubmitting && registerError === '');
+  const onError = Boolean(!isFormSubmitting && registerError);
```
wlechowicz, line 28: Booleans, so names should start with `is` or `has`.

wlechowicz (changes requested):
Good improvements from last review! Couple of things I spotted are inline.

app/components/Form/ServerSideMessage/index.js
marcelina-hasiak: Thank you for this one. I should have read SC documentation more :|

app/components/Form/UserSignUpForm/index.js
wlechowicz: It is not and sign-in needs fixing if it does that.

app/components/Form/ServerSideMessage/ServerSideMessage.stories.js
```diff
@@ -0,0 +1,20 @@
...
+  component: ServerSideMessage,
+};
+
+const Template = ({ children, ...args }) => {
+  return <ServerSideMessage {...args}> {children} </ServerSideMessage>;
+};
+
+export const Default = Template.bind({});
+Default.args = {
+  errorType: false,
```
wlechowicz, line 13: Forgot to change prop names in story.

wlechowicz (changes requested):

wlechowicz (approved):

hiimsagan (approved):

PR #94, opened 2022-03-25
feat: my list backend #1734 

rast-mk1:


wlechowicz:
> create client (like services/movieDb.js) 

Yes, pages will need to use a method, rather than an API endpoin, to fetch My List data.

pages/api/mylist/[movieId].js
```diff
@@ -0,0 +1,38 @@
...
+      db.collection('user_lists').updateOne(query, command, options);
+      return res.status(201).json({});
+    } else {
+      return res.status(200).json({});
+    }
+  }
+
+  // different approach with implementation of Array.filter would be nice
+  if (req.method == 'DELETE') {
+    const command = { $pull: { items: movieId } };
```
Comment (wlechowicz) [inline pages/api/mylist/[movieId].js:31]: If this works, then it's better, since it's a single operation.
With Array.filter you'd need to get the data, modify it and save back to DB with 2 operations.

pages/api/mylist/[movieId].js
```diff
@@ -0,0 +1,38 @@
...
+  const client = await mongoClient;
+  const db = client.db(process.env.MONGODB_DB);
+
+  const { movieId } = req.query;
+  const userId = req.session.id;
+  const query = { _id: userId };
+
+  // different approach with implementation of Set would be nice
+  if (req.method == 'PUT' && movieId > 0) {
+    const command = { $addToSet: { items: movieId } };
```
Comment (wlechowicz) [inline pages/api/mylist/[movieId].js:14]: This is better than Set in JS, since it's a single operation instead of get-modify-save.
Only issue is 
> [$addToSet](https://www.mongodb.com/docs/manual/reference/operator/update/addToSet/#mongodb-update-up.-addToSet) does not guarantee a particular ordering of elements in the modified set.

which might mean we don't preserve order of My List. See if it's true, if it preserves the order then keep it.

wlechowicz (changes requested):

pages/api/mylist/[movieId].js
Comment (rast-mk1) [reply pages/api/mylist/[movieId].js:14]: Did some test using few users and seems like order doesn't change.

pages/api/mylist.js
```diff
@@ -0,0 +1,16 @@
+import mongoClient from 'server/mongoDb';
+import withAuth from 'server/withAuth';
+
+async function handler(req, res) {
+  const client = await mongoClient;
+  const db = client.db(process.env.MONGODB_DB);
+
+  if (req.method == 'GET') {
+    const list = await db.collection('user_lists').findOne({ _id: req.session.id });
```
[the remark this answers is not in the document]
rast-mk1: I did that on purpose, to have rather one `_id` than auto generated `_id` and then `user_id` as well. Should I have both `_id` and `user_id`?

pages/api/mylist.js
```diff
@@ -0,0 +1,16 @@
...
+    if (!list) {
+      return res.status(200).json([]);
+    }
+    return res.status(200).json(list?.items);
```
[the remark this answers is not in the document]
rast-mk1: Got it.

pages/api/mylist/[movieId].js
```diff
@@ -0,0 +1,38 @@
...
+    const options = { upsert: true };
+    const userList = await db.collection('user_lists').findOne({ _id: req.session.id });
```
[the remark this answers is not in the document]
Comment (rast-mk1) [reply pages/api/mylist/[movieId].js:16]: Oversight. Will fix that.

pages/api/mylist/[movieId].js
```diff
@@ -0,0 +1,38 @@
...
+    const numberOfMoviesOnList = userList?.items.length;
+    if (!userList && movieId.length > 0 && movieId.length < 50) {
```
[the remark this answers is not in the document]
Comment (rast-mk1) [reply pages/api/mylist/[movieId].js:18]: It does work like intended, though. Will look into that.

pages/api/mylist/[movieId].js
```diff
@@ -0,0 +1,38 @@
...
+      db.collection('user_lists').updateOne(query, command, options);
+      return res.status(201).json({});
+    } else if (movieId.length > 0 && movieId.length < 50 && numberOfMoviesOnList < 50) {
```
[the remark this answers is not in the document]
Comment (rast-mk1) [reply pages/api/mylist/[movieId].js:21]: I did that, because it worked fine for existing list, but I was getting HTTP 500 if there was no list (no `movieId.length`).

pages/api/mylist/[movieId].js
Comment (rast-mk1) [reply pages/api/mylist/[movieId].js:31]: It works.

pages/api/mylist.js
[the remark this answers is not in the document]
wlechowicz: While we can use `_id`s for document relations between collections (and consider `user.id` redundant), we should keep them autogenerated as ObjectIds. But then we need to change what we return from sign in.
Also, using `user.id` as `_id` in a different collection doesn't sound like a good idea, since it will not scale - day will come when you need to add another collection related to user and you can't reuse the same logic because it would lead to duplicate `_id`s. Correct me if I'm wrong, but `_id` must be unique to the document across all collections in a database and we should express relationships between collections as a secondary field in the document. Unless you want to nest documents in other documents (like a document contains a list of child-documents) - then you would use `_id`.
I think using `user.id` generated by us makes it more DB-agnostic and lets the engine manage document id uniqueness.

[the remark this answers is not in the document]
rast-mk1: Okay, that makes sense and having another `id` won't really complicate anything, so I am for it.

pages/api/mylist/[movieId].js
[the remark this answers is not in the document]
Comment (wlechowicz) [reply pages/api/mylist/[movieId].js:21]: `movieId` is not the list, it's the input param from query.

pages/api/mylist/[movieId].js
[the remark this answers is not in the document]
Comment (wlechowicz) [reply pages/api/mylist/[movieId].js:18]: It cannot work like intended, because `movieId` is not the list, it's the input param from query, so it's a numerical string between 0 and something so the length will always be between 0 and 50.

docs/apiDocs.yml
```diff
@@ -218,6 +220,92 @@ paths:
...
+            application/json:
+              schema:
+                type: array
+
+  /api/mylist/{movieId}:
+    put:
+      tags:
+        - my list
+      summary: Insert movie into user's movie list.
+      description: Insert movie into user's movie list.
```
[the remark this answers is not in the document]
rast-mk1: Changed summary and description like suggested.

pages/api/mylist/[movieId].js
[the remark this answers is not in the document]
Comment (rast-mk1) [reply pages/api/mylist/[movieId].js:18]: Sorry, I got things mixed up a bit. I am limiting `movieId` length as suggested in improvements task from backlog: https://tracker.intive.com/jira/browse/P2022-1870 The goal is not to handle inputs with 50 or more digits, but I am willing to take different approach or remove this limit if needed.

pages/api/mylist/[movieId].js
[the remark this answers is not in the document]
Comment (rast-mk1) [reply pages/api/mylist/[movieId].js:21]: Yes, my bad. I meant `numberOfMoviesOnList ` not `movieId.length`,

pages/api/mylist/[movieId].js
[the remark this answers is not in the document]
Comment (wlechowicz) [reply pages/api/mylist/[movieId].js]: Well the length limitation is for string params, this is a number and a number that has 50 digits is very well over the id limit we actually need.

app/server/services/myListClient.js
```diff
@@ -0,0 +1,21 @@
+import mongoClient from 'server/mongoDb';
+import { getMovieById } from './movieDb';
+
+/**
+ * Get user's movie list
+ *
+ * @export
+ * @param {string|number} req Request containing user's session id
```
wlechowicz, line 8: `req` is neither a string nor a number.

app/server/services/myListClient.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+export async function getMyList(req) {
+  const client = await mongoClient;
+  const db = client.db(process.env.MONGODB_DB);
+  const userList = await db.collection('user_lists').findOne({ user_id: req.session.id });
+
+  let myList = new Array();
+  for (let index = 0; index < userList.items.length; index++) {
+    myList.push(await getMovieById(userList.items[index]));
+  }
```
wlechowicz, line 19: Use `Array.map` on `userList` for this, but make sure it's a valid result first.

docs/apiDocs.yml
```diff
@@ -218,6 +220,77 @@ paths:
...
+      summary: Add movie to user's movie list.
+      description: Add movie to user's movie list.
+      parameters:
+        - in: path
+          name: movieId
+          required: true
+          description: Id of a movie to add to user's movie list.
+          schema:
+            type: integer
+      security: []
```
wlechowicz, line 250: Not right, it is secured with auth.

docs/apiDocs.yml
```diff
@@ -218,6 +220,77 @@ paths:
...
+      responses:
+        200:
+          description: ""
```
wlechowicz, line 253: Why empty?

docs/apiDocs.yml
```diff
@@ -218,6 +220,77 @@ paths:
...
+      summary: Delete movie from user's movie list.
+      description: Delete movie from user's movie list.
+      parameters:
+        - in: path
+          name: movieId
+          required: true
+          description: Id of a movie to delete from user's list of movies.
+          schema:
+            type: integer
+      security: []
```
wlechowicz, line 277: Nope, this is secure as well.

docs/apiDocs.yml
```diff
@@ -218,6 +220,77 @@ paths:
...
+      responses:
+        200:
+          description: ""
```
wlechowicz, line 280: Empty.

pages/api/mylist/[movieId].js
```diff
@@ -0,0 +1,42 @@
...
+
+  const userId = req.session.id;
+  const query = { user_id: userId };
+
+  if (req.method === 'PUT') {
+    const command = { $addToSet: { items: id } };
+    const options = { upsert: true };
+    const userList = await db.collection('user_lists').findOne(query);
+    const numberOfMoviesOnList = userList?.items.length;
+    if (Number.isInteger(id) && id > 0) {
```
Comment (wlechowicz) [inline pages/api/mylist/[movieId].js:18]: Duplicated. Also should be at the very top of the handler - input params validation comes first, if `id` isn't legit then why fetch data from DB etc. If `id` is bad, return 400.

pages/api/mylist/[movieId].js
```diff
@@ -0,0 +1,42 @@
...
+      if (!userList) {
+        db.collection('user_lists').updateOne(query, command, options);
+        return res.status(200).json();
+      } else if (numberOfMoviesOnList < 50) {
+        db.collection('user_lists').updateOne(query, command, options);
+        return res.status(200).json();
```
Comment (wlechowicz) [inline pages/api/mylist/[movieId].js:24]: This if and the else to the same exact thing.

wlechowicz (changes requested):

app/server/services/myListClient.js
rast-mk1: Meaning, that movie with given id exists in theMovieDB?

app/server/services/myListClient.js
```diff
@@ -0,0 +1,22 @@
...
+  const client = await mongoClient;
+  const db = client.db(process.env.MONGODB_DB);
+  const userList = await db.collection('user_lists').findOne({ user_id: req.session.id });
+
+  if (Array.isArray(userList.items)) {
+    const myList = await Promise.all(userList.items.map((id) => getMovieById(id)));
+    const myFilteredList = myList.filter((element) => element !== null);
+    return myFilteredList;
+  } else {
+    return;
```
wlechowicz, line 20: `return [];` because it will be used not only through endpoint by from server-side component code as well.

pages/api/mylist/[movieId].js
```diff
@@ -0,0 +1,36 @@
+import mongoClient from 'server/mongoDb';
+import withAuth from 'server/withAuth';
+async function handler(req, res) {
+  if (!['put', 'delete'].includes(req.method.toLowerCase())) {
+    return res.status(405).send();
+  }
```
Comment (wlechowicz) [inline pages/api/mylist/[movieId].js:6]:
```suggestion
  const method = req.method.toLowerCase();
  if (!['put', 'delete'].includes(method)) {
    return res.status(405).send();
  }
```
and use `method` in the `if`s later on, for consistency.

wlechowicz (changes requested):

wlechowicz (approved):

PR #95, opened 2022-03-27
feat:  user action menu #1729

wolnio:


app/components/ActionMenu/index.js
```diff
@@ -0,0 +1,40 @@
+import { Wrapper, UserInfo, ActionList, ActionBtn } from './ActionMenu.styles';
+import Button from '../UI/Button';
+import { useSession as useSessionActual } from 'next-auth/react';
```
weronika299, line 3: I think that this import (line 3) should be the first on the list.

weronika299 (changes requested):

app/components/ActionMenu/index.js
```diff
@@ -0,0 +1,40 @@
...
+  const state = useActions({ isModalOpen: false, content: '' });
+  const { isModalOpen, content } = state;
+
+  return (
+    <>
+      <Wrapper>
+        <UserInfo>
+          {session?.user ? (
+            <>
+              <UserAvatar size={100} />
```
wlechowicz, line 20: You need to pass in `avatar` prop, avatar is blank otherwise.

app/components/ActionMenu/ActionMenu.styles.js
```diff
@@ -0,0 +1,44 @@
+import Button from 'components/UI/Button';
+import styled from 'styled-components';
+
+const Wrapper = styled.div`
+  padding: 32px 16px;
+  width: 300px;
+  height: 100%;
```
wlechowicz, line 7: 100% of what? The outer-most element shouldn't have a set height - instead, it's the content+paddings+margins that will determine its height from inside naturally.

app/components/ActionMenu/ActionMenu.styles.js
```diff
@@ -0,0 +1,44 @@
...
+  border-radius: 8px;
+`;
+
+const UserInfo = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  font-size: 1.25em;
+
+  .username {
```
wlechowicz, line 19: Unexpected use of classes. Use a separate styled component instead.

app/components/ActionMenu/ActionMenu.styles.js
```diff
@@ -0,0 +1,44 @@
...
+    margin: 16px 0;
+  }
+`;
+
+const ActionList = styled.ul`
```
wlechowicz, line 24: `ul` should have list items `li` inside, you have just buttons. If you want to have this as a list (which isn't a bad idea) then you need to wrap buttons in list items.

app/components/ActionMenu/index.js
```diff
@@ -0,0 +1,40 @@
...
+            <>Not signed in :(</>
+          )}
+        </UserInfo>
+        <ActionList>
+          <ActionBtn onlyIcon fullWidth onClick={() => openModal(<h2>Account</h2>)}>
+            Account
+            <Icon size="lg" rotation={270} type="arrow-down" />
+          </ActionBtn>
+        </ActionList>
+        {isModalOpen && <Modal content={content} />}
```
wlechowicz, line 33: Modal cannot be a part of this component.

app/components/ActionMenu/index.js
```diff
@@ -0,0 +1,40 @@
+import { Wrapper, UserInfo, ActionList, ActionBtn } from './ActionMenu.styles';
+import Button from '../UI/Button';
+import { useSession as useSessionActual } from 'next-auth/react';
+import UserAvatar from 'components/UI/UserAvatar';
+import Icon from '../UI/Icon';
+import Modal from 'components/Modal';
+import { useActions, openModal } from 'actions/app';
+
+export default function ActionMenu({ isOpen, useSession = useSessionActual }) {
```
wlechowicz, line 9: `isOpen` is not used. Accept `user` as prop instead of using session.
Use existing `openModal` action from app actions, you can mock it in story template (use isModalOpen in story Template to show a modal).

app/components/ActionMenu/index.js
```diff
@@ -0,0 +1,40 @@
...
+  const { data: session } = useSession();
+  const state = useActions({ isModalOpen: false, content: '' });
+  const { isModalOpen, content } = state;
```
wlechowicz, line 12: Use this only in story for presentation, not in the component itself.

app/components/ActionMenu/index.js
```diff
@@ -0,0 +1,40 @@
...
+          )}
+        </UserInfo>
+        <ActionList>
+          <ActionBtn onlyIcon fullWidth onClick={() => openModal(<h2>Account</h2>)}>
+            Account
+            <Icon size="lg" rotation={270} type="arrow-down" />
+          </ActionBtn>
+        </ActionList>
+        {isModalOpen && <Modal content={content} />}
+        <Button fullWidth href="#">
```
wlechowicz, line 34: `onClick={signOut}` Check app\components\Account\index.js for reference.

wlechowicz (changes requested):
This menu will be controlled by Header, so rather than ActionMenu using useSession, it should accept `user` object in props, that will get passed down from Header.
You can use modal to showcase feature in Storybook, but not in the actual component, because there already is a real modal on the page.

app/components/ActionMenu/index.js
wolnio: Do you mean to create separate story where isModalOpen is set to true so modal is already open when we open that story?

wlechowicz (approved):

app/components/ActionMenu/index.js
wlechowicz: I meant you could make a story where you put the menu on the left and the modal opens with "Account" when you press the Account button in your menu. To do that, the story Template component would have to host the modal like app does and the action from menu would open that modal. Not worth IMO since it's not testing the account, but just the button.

weronika299 (approved):

wlechowicz:
One more request, make the menu usable - make it open and close from the Header.
Opens when user avatar in Header is clicked (make Header's UserAvatar a button) and closes either when avatar is clicked again or with a useOnClickOutside hook for a click outside of the menu container.

app/components/ActionMenu/index.js
```diff
@@ -0,0 +1,27 @@
+import { Wrapper, UserInfo, Username, ActionList, ListItem, ActionBtn } from './ActionMenu.styles';
+import Button from '../UI/Button';
+import UserAvatar from 'components/UI/UserAvatar';
+import Icon from '../UI/Icon';
+import { signOut } from 'next-auth/react';
```
weronika299, line 5: One more thing, It could be first on the list.

hiimsagan (approved):

marcelina-hasiak:
Looks good for me too. Only one thing you need to adjust is Header story.
Zrzut ekranu 2022-04-5 o 15 51 55

PR #96, opened 2022-03-27
feat: home page #1728

weronika299:


app/components/Pages/HomePage/HomePage.styles.js
```diff
@@ -0,0 +1,47 @@
...
+  margin-top: -25vh;
+  color: #fff;
+  ${device.tablet} {
+    margin-top: -35vh;
+  }
+`;
+
+export const CarouselHeader = styled.h5`
+  margin-left: 30px;
+  height: 0;
```
wolnio, line 26: Positioning of a header looks strange. At the end you get what you want, but the text isn't even in the scope of it's container - it can be easily seen on the image below, text is inside of it's padding. I don't think `height: 0` for text field is good solution. It's good to remember that headers h1 - h6 have build in `margin top and bottom`. For h5 margin equals to something around `1.67em`.
image

wolnio (changes requested):

app/components/Pages/HomePage/HomePage.styles.js
weronika299: Yes, I will change this to positioning  of each CarouselWrapper. Thanks!

rast-mk1 (changes requested):
For some reason, footer is gone and OpenModal button is there instead.
obraz
Also, you could add padding around MovieCard or make background transparent.

pages/index.js
```diff
@@ -1,5 +1,60 @@
...
+
+  return (
+    <>
+      {session?.user ? (
+        <HomePage listOfCategories={listOfCategories} heroMovie={heroMovie} />
+      ) : (
+        <>Not signed in :(</>
+      )}
+      <Button onClick={() => openModal(<h2>Modal</h2>)}>OpenModal</Button>
+      {isModalOpen && <Modal content={content} />}
```
wlechowicz, line 26: Delete this please.

pages/index.js
```diff
@@ -1,5 +1,60 @@
...
+    let listOfCategories = await Promise.all(
+      genresList.map(async (category) => {
+        const list = await getListOfMoviesByCategoryId(category.id);
+        const listOfMovies = list.results;
+        return { ...category, listOfMovies };
+      })
+    );
+
+    const listOfHeroMoviesId = [508947, 634649, 696806, 568124, 414906];
+    const chosenMovieId = listOfHeroMoviesId[Math.floor(Math.random() * listOfHeroMoviesId.length)];
```
wlechowicz, line 43: I'd call it `randomMovieId`

pages/index.js
```diff
@@ -1,5 +1,60 @@
...
+    const heroMovie = await getMovieById(chosenMovieId);
+    const popularList = await getListOfPopularMovies(req);
+    const popularCategory = { id: 1, name: 'Popular', listOfMovies: popularList.results };
+    listOfCategories.unshift(popularCategory);
+
+    return {
+      props: {
+        heroMovie,
+        listOfCategories,
+        pageLayout: { header: true, footer: false },
```
wlechowicz, line 54: Let's have the footer on, this is the page that user will spend most of their time on (apart from the player, of course).

pages/index.js
```diff
@@ -1,5 +1,60 @@
...
+      <Button onClick={() => openModal(<h2>Modal</h2>)}>OpenModal</Button>
+      {isModalOpen && <Modal content={content} />}
+    </>
+  );
+}
+
+export async function getServerSideProps({ req }) {
+  try {
+    let Promise = require('promise');
+    let listOfCategories = await Promise.all(
```
wlechowicz, line 34: Make this a const, and then instead of unshifting popular list onto this, return it as `[popularCategory, ...listOfCategories]`.

package.json
```diff
@@ -24,6 +24,7 @@
     "next": "12.1.0",
     "next-auth": "^4.2.1",
     "node-fetch": "^2.6.7",
+    "promise": "^8.1.0",
```
wlechowicz, line 27: Nope.

app/components/Pages/HomePage/index.js
```diff
@@ -0,0 +1,32 @@
+import {
+  HomePageWrapper,
+  CarouselHeader,
+  CarouselWrapper,
+  HomePageFooter,
+  MovieHeroWrapper,
+} from './HomePage.styles';
+
+import Carousel from 'components/Carousel';
+import Link from 'next/link';
```
wlechowicz, line 10: Imports from outside of our app (packages etc.) should be the first on the list, then out own imports. Move this to the top.

app/components/Pages/HomePage/index.js
```diff
@@ -0,0 +1,32 @@
...
+export default function HomePage({ listOfCategories, heroMovie }) {
+  return (
+    <HomePageWrapper>
+      <MovieHeroWrapper>
+        <MovieHero movieData={heroMovie} />
+      </MovieHeroWrapper>
+      <CarouselWrapper>
+        {listOfCategories.map((list) => (
+          <div key={list.id}>
+            <Link href="/#" passHref>
```
wlechowicz, line 22: Please use correct links for this and the movie card. This would be `/category/${id}`

app/components/Pages/HomePage/index.js
```diff
@@ -0,0 +1,32 @@
...
+import MovieHero from 'components/MovieHero';
+
+export default function HomePage({ listOfCategories, heroMovie }) {
+  return (
+    <HomePageWrapper>
+      <MovieHeroWrapper>
+        <MovieHero movieData={heroMovie} />
+      </MovieHeroWrapper>
+      <CarouselWrapper>
+        {listOfCategories.map((list) => (
```
wlechowicz, line 20: Destructure props from `list` here.

app/components/Pages/HomePage/index.js
```diff
@@ -0,0 +1,32 @@
...
+import Link from 'next/link';
+import MovieHero from 'components/MovieHero';
+
+export default function HomePage({ listOfCategories, heroMovie }) {
+  return (
+    <HomePageWrapper>
+      <MovieHeroWrapper>
+        <MovieHero movieData={heroMovie} />
+      </MovieHeroWrapper>
+      <CarouselWrapper>
```
wlechowicz, line 19: Carousels are left-aligned, should be centered.

app/components/Carousel/index.js
```diff
@@ -67,19 +68,25 @@ const Carousel = ({ movies = [] }) => {
...
     <CarouselOuter>
       <CarouselInner activeScreen={activeScreen}>
-        {movies.map(({ id, title }) => (
-          <CarouselItem key={id} title={title} tileCount={tilesPerScreen} />
+        {movies.map(({ id, title, images }) => (
+          <CarouselItem
+            key={id}
+            title={title}
+            tileCount={tilesPerScreen}
+            src={images?.poster?.original}
```
wlechowicz, line 87: MovieCard should not be using the original image size, but a smaller one, like M or L.

app/components/Carousel/index.js
```diff
@@ -67,19 +68,25 @@ const Carousel = ({ movies = [] }) => {
     setActiveScreen(newIndex);
   };
 
-  const CarouselItem = ({ tileCount }) => (
+  const CarouselItem = ({ tileCount, src, title, movieId }) => (
```
wlechowicz, line 71:
```suggestion
  const CarouselItem = ({ tileCount, ...props }) => (
```

app/components/Carousel/index.js
```diff
@@ -67,19 +68,25 @@ const Carousel = ({ movies = [] }) => {
...
     <CarouselItemWrapper tileCount={tileCount}>
       <CarouselItemInner>
-        <div style={{ height: '320px' }}>MovieCard will be here</div>
+        <MovieCard src={src} title={title} movieId={movieId} />
```
wlechowicz, line 74:
```suggestion
        <MovieCard {...props} />
```

wlechowicz (changes requested):
Grey background behind MovieCard shouldn't be there, but that's probably Carousel bug. Movie title in carousels should be left aligned, not centered - that's probably a MovieCard bug.

pages/index.js
```diff
@@ -1,5 +1,60 @@
...
+      genresList.map(async (category) => {
+        const list = await getListOfMoviesByCategoryId(category.id);
+        const listOfMovies = list.results;
+        return { ...category, listOfMovies };
+      })
+    );
+
+    const listOfHeroMoviesId = [508947, 634649, 696806, 568124, 414906];
```
wlechowicz, line 42: I just had this idea that, instead of this being a random movie out of a list of hand-piked movies, it would be cool if it was a random movie from the 10 most popular movies -  you have the popular list here already, so it should be easy, like `const randomIdFromMostPopularTop10 = popularList[Math.floor(Math.random() * Math.min(10, popularList.length)]`.
This way the possible heroes will change over time and will stay relevant.

app/components/UI/MovieCard/MovieCard.styles.js
```diff
@@ -20,19 +19,19 @@ export const PosterWrapper = styled(flexDivStyle)`
 export const Wrapper = styled(flexDivStyle)`
   flex-direction: column;
   align-items: flex-start;
-
+  aspect-ratio: 1 / 1.8;
```
wlechowicz, line 22: This is added to the wrong element. It's on the whole wrapper instead of only the poster image (PoterWrapper). It makes the active area of the `a` extend down below the title element.
Then, 1.8 denominator is too much and it makes the image too narrow, clipping titles. `1 / 1.5` seems to be the sweet spot.

pages/index.js
weronika299: Okay but this is our old footer that we used before design, so should i change the old footer to the new one (does the new one only have height and color?) in the footer component and then make it visible?

rast-mk1 (changes requested):
When you press chevron to get another portion of movies from Carousel, images are not loaded yet. Need to load remaining images before user clicks chevron. Also, images that are on first slide are still loading when page is rendered.

weronika299:
> When you press chevron to get another portion of movies from Carousel, images are not loaded yet. Need to load remaining images before user clicks chevron. Also, images that are on first slide are still loading when page is rendered. EDIT: inside `MoviePoster` component, there is `blurDataURL={'images/placeholder-blur.png'}` that should blur images, when loading. It used to work fine when I tested it, but now it doesn't. I will try to look into that.

pages/index.js
wlechowicz: Keep the footer as is, changing it is not in the scope of this story.

weronika299: Okay :)

wlechowicz:
> When you press chevron to get another portion of movies from Carousel, images are not loaded yet. Need to load remaining images before user clicks chevron. Also, images that are on first slide are still loading when page is rendered. EDIT: inside `MoviePoster` component, there is `blurDataURL={'images/placeholder-blur.png'}` that should blur images, when loading. It used to work fine when I tested it, but now it doesn't. I will try to look into that.

The images loading in late is the behavior of next/image as it doesn't preload images that aren't visible. It's a good optimization since we don't need to load in all of the 100+ images in the page at once. Changing the image used from original size poster to a smaller image should greatly help the load ins.

pages/index.js
```diff
@@ -1,5 +1,51 @@
...
+        return { ...category, listOfMovies };
+      })
+    );
+
+    const popularList = await getListOfPopularMovies(req);
+    const resultsOfPopularList = popularList.results;
+    const popularCategory = { id: 1, name: 'Popular', listOfMovies: popularList.results };
+    const listOfCarousels = [popularCategory, ...listOfCategories];
+    const randomIdFromMostPopularTop10 =
+      resultsOfPopularList[Math.floor(Math.random() * Math.min(10, resultsOfPopularList.length))]
```
marcelina-hasiak, line 38: I would move this math operation to utils folder and wrap with function. It seems to me we have plenty of Math.blablabla and some of them maybe we can reuse in the future. Not to mention it would be nicer to read the code here.

marcelina-hasiak (changes requested):
Some little suggestions from my side. 
Apart from that I really like the outcome :) Good job!!!

app/components/Pages/HomePage/HomePage.styles.js
```diff
@@ -0,0 +1,45 @@
...
+  cursor: pointer;
+  ${device.tablet} {
+    font-size: 1.25rem;
+    font-weight: 700;
+  }
+
+  &:hover {
+    text-decoration: underline;
+  }
+`;
```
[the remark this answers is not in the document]
weronika299: Yes, we have already 'to do' tasks on Jira.

pages/index.js
wlechowicz: I'd drop a comment on line 37 to explain what it does, but it won't be used anywhere else so let it sit here.

app/components/UI/MovieCard/index.js
```diff
@@ -3,18 +3,17 @@ import Link from 'next/link';
 import MoviePoster from 'components/UI/MoviePoster';
 import { Wrapper, PosterWrapper, TitleWrapper, Title } from './MovieCard.styles';
 
-export default function MovieCard({ src, title, alt, href, ...props }) {
-  /* TODO placeholder for link to specific movie  */
-  href = '#';
+export default function MovieCard({ ...props }) {
```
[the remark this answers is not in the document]
wlechowicz: If anything, it should be `MovieCard({ movie })` since it's kinda stupid to pass props one by one when they come from the same object.

wlechowicz (approved):
Minor issues discussed in Marcelina's review. One improvement that would make it nicer would be if the header was being hidden with an animation, rather than this BAM gone BAM visible.

Overall very nice, Let's get this in.

rast-mk1 (approved):

rast-mk1:
Starting to look like a real thing :)

app/components/UI/MovieCard/index.js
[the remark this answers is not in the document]
weronika299: So is my solution also correct or should it be changed?

app/components/Header/index.js
```diff
@@ -18,41 +18,52 @@ import SearchMoviesInput from 'components/UI/SearchMoviesInput';
...
+    } else {
+      setIsCarouselSection(true);
+    }
+  };
   return (
     <>
-      <NavigationBar>
-        <MainPanel>
+      {!isCarouselSection ? (
+        <NavigationBar remove={remove}>
```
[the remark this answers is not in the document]
weronika299: This prop is unnecessary. I think I forgot to delete it after trying a different solution.

app/components/UI/MovieCard/index.js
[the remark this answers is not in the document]
marcelina-hasiak: Of course it is correct. We just use destructure where it's possible to minimize amount of code and maximize readability.

marcelina-hasiak (approved):

wlechowicz (approved):

wlechowicz:
Let's get this in.

wlechowicz:
Dismissing one stale review in light of the owner not being active. You can merge.

PR #97, opened 2022-03-28
fix: authcontainer backgroud size #1784

weronika299:


wlechowicz (changes requested):
- blur doesn't seem to work
- you shouldn't need to set blurDataUrl manually to files that we host statically, Image should do it itself (https://nextjs.org/docs/api-reference/next/image#placeholder)
- you might have to `import` the image like this https://nextjs.org/docs/basic-features/image-optimization#local-images
- the new image is 1.1 MB yet the quality on the page is poor due to Image applying optimizations I assume. Default `quality` is 75, try setting it to 90 - 100.

rast-mk1:
> * blur doesn't seem to work
Actually, it does work: 
obraz

app/components/Pages/AuthPageContainer/index.js
```diff
@@ -1,11 +1,20 @@
...
   return (
     <Wrapper>
+      <AuthBackground
+        src={'/images/hero-background-image.jpg'}
+        alt="background-image"
+        layout={'fill'}
+        objectFit={'cover'}
+        objectPosition={'top left'}
+        placeholder="blur"
+        blurDataURL={'/images/blur-auth-placeholder.png'}
```
rast-mk1, line 16: You don't really need `blurDataURL` for your purpose, but if you want to use it, image should be much smaller

> It will be enlarged and blurred, so a very small image (10px or less) is recommended. 

You could use `public/images/placeholder-blur.png` for that.

app/components/Pages/AuthPageContainer/AuthPageContainer.styles.js
```diff
@@ -10,10 +12,13 @@ export const Wrapper = styled.div`
...
-    background-image: url('images/hero-background-image.jpg');
-    background-position: top left;
-    background-repeat: no-repeat;
-    background-size: cover;
+  }
+`;
+
+export const AuthBackground = styled(Image)`
+  display: none !important;
+  ${device.tablet} {
```
marcelina-hasiak, line 20: device.mobile ?

app/components/Pages/AuthPageContainer/index.js
marcelina-hasiak: and 615kb for placeholder - I think it's still too heavy

marcelina-hasiak (changes requested):

app/components/Pages/AuthPageContainer/AuthPageContainer.styles.js
weronika299: Why?

marcelina-hasiak: I have desktop first approach at my job and I forgot we've got here just the opposite; sorry for mixing it up ;)

marcelina-hasiak (approved):

marcelina-hasiak:
I approved PR but I forgot  to add that I couldn't recreate blur image on storybook. Is it already fixed?

weronika299:
> I approved PR but I forgot to add that I couldn't recreate blur image on storybook. Is it already fixed?

Unfortunately I haven't fixed it yet. :/

wlechowicz (approved):
Works well with throttled network and disabled cache. If blur doesn't work in Storybook, well too bad for Storybook.

rast-mk1 (approved):

PR #98, opened 2022-03-30
fix: minor fixes to Carousel and MovieCard #2003 #2004 #2005

rast-mk1:
How it looks after making changes according to task instructions:

obraz
obraz


wlechowicz (changes requested):
Title and background are fixed. 
Nav buttons don't look good with 100% because they stick out at the bottom. Maybe something like 80% but centered vertically would work better. They also need to be a bit more opaque, they are hardly visible now.

wlechowicz (approved):

PR #103, opened 2022-03-31
feat: logic for SearchInput #1730

marcelina-hasiak:


app/components/UI/MovieGrid/index.js
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <StyledMovieGrid>
+      {listOfMovies.map(({ id, title, images }) => (
+        <MovieCard
+          ratio="2/1"
+          key={id}
+          id={id}
+          title={title}
+          src={images?.backdrop?.l || images?.poster?.l}
+        />
```
rast-mk1, line 15: IMO, having only `images.poster` and vertical `MovieCards` would be simpler and look better.

wlechowicz (changes requested):
Code looks good, but please revert the MovieCard changes and use vertical posters in the grid. The backdrop with tiny title makes it hard to know what the result is.

wlechowicz (approved):

weronika299 (approved):

PR #107, opened 2022-04-02
feat: movie page

hiimsagan:


hiimsagan:
Waiting for movie details component

pages/movie/[id].js
```diff
@@ -0,0 +1,47 @@
...
+        <div>helo</div>
+      ) : (
+        <>Not signed in :(</>
+      )}
+    </>
+  );
+}
+
+export async function getServerSideProps(context) {
+  const { id } = context.query;
```
Comment (wlechowicz) [inline pages/movie/[id].js:31]:
```suggestion
  const { id } = context.params;
```
Also you can destructure `params` from context in the function declaration.

pages/movie/[id].js
```diff
@@ -0,0 +1,47 @@
...
+
+  const movie = await getMovieById(id);
+  const movieData = movie;
```
Comment (wlechowicz) [inline pages/movie/[id].js:34]: What's the point of this?

pages/movie/[id].js
```diff
@@ -0,0 +1,47 @@
...
+  const SimilarMovies = await getListOfSimilarMoviesById(id);
+  const listOfSimilarMovies = SimilarMovies.results;
+  const RecommendedMovies = await getListOfRecommendedMoviesById(id);
+  const ListOfRecommendedMovies = RecommendedMovies.results;
```
Comment (wlechowicz) [inline pages/movie/[id].js:38]: Why the mix of camelCase and PascalCase? Why not desctructure `results` right from the response?
Also all of theses should be inside a try..catch with a fallback to empty data if something blows up.

pages/movie/[id].js
```diff
@@ -0,0 +1,45 @@
...
+
+export async function getServerSideProps(context) {
+  const { id } = context.params;
+  try {
+    const movieData = await getMovieById(id);
+
+    const listOfSimilarMovies = await getListOfSimilarMoviesById(id).then((data) => data.results);
+
+    const listOfRecommendedMovies = await getListOfRecommendedMoviesById(id).then(
+      (data) => data.results
```
Comment (marcelina-hasiak) [inline pages/movie/[id].js:32]: What about Promise.all?  You'd save some time between requests.

marcelina-hasiak (changes requested):

PR #108, opened 2022-04-02
fix: clickable buttons in mobile view #2094

weronika299:


hiimsagan (approved):
Is woorking, but btw sign-up should be center on mobiles 
Zrzut ekranu 2022-04-02 225055
?

app/components/Pages/AuthPageContainer/AuthPageContainer.styles.js
```diff
@@ -9,16 +8,16 @@ export const Wrapper = styled.div`
   width: 100vw;
   background-color: #2a2535;
   color: white;
```
lenasokolova, line 10: I believe it's better to use #fff instead of words like "white" but it's a little thing :)

wlechowicz (changes requested):
I can't accept this.
1. the span you're trying to affect is part of next/Image, therefore this is a hack, not a fix
2. we should never need `!important` with styled-components, another sign you're trying to do something fishy

Did you try the alternative solution I proposed, with just a background-image instead of next/Image and using two images, small blur and the actual image at once in single background-image statement?
`background-image: url('actual-image'), url('blur-image');` it goes top to bottom from left to right, so the first one will be on top, so the blur will show through as the actual image is loading.

weronika299:
> I can't accept this.
> 
> 1. the span you're trying to affect is part of next/Image, therefore this is a hack, not a fix
> 2. we should never need `!important` with styled-components, another sign you're trying to do something fishy
> 
> Did you try the alternative solution I proposed, with just a background-image instead of next/Image and using two images, small blur and the actual image at once in single background-image statement? `background-image: url('actual-image'), url('blur-image');` it goes top to bottom from left to right, so the first one will be on top, so the blur will show through as the actual image is loading.

Okay, you're right. Yes, I tried this solution, but I think I did something wrong because it didn't work for me, but I will try again. Thank you!

wlechowicz (approved):
Let's leave it as is.
Doesn't seem to work for me, the placeholder image shows up only after the big one is downloaded (disabled caching with throttled network). Perhaps it would have to be split into 2 separate containers, one on top of the other. Or maybe just converting the JPEG to a progressive one would fix this once and for all.

PR #110, opened 2022-04-03
fix: fix sign up process #2088

marcelina-hasiak:


wlechowicz:
Issue was Vercel/NextJS setup, in production env the domain is static, but VERCEL_URL env variable was pointing to the deployment URL (different for each build), not the production domain, so the origin was wrong, hence CORS error.

PR #111, opened 2022-04-04
feat: add mylist carousel #1902

weronika299:


rast-mk1 (approved):
Looks good. Not sure, if carousel for `myList` should have title "MyList" but rather"My List" or something else

weronika299:
> Looks good. Not sure, if carousel for `myList` should have title "MyList" but rather"My List" or something else

Yes, you're right. I didn't notice it. I will corect it. Thanks!

rast-mk1:
One more thing: probably would be better, if my List is not rendered with 0 MovieCards when it is empty: 
obraz

lenasokolova (approved):

pages/index.js
```diff
@@ -30,10 +32,13 @@ export async function getServerSideProps({ req }) {
       })
     );
 
+    const session = await verifyJwtInRequest(req);
+    const myList = await getMyList({ session });
```
wlechowicz, line 36: You shouldn't do this here, `getMyList` should pull the session from req on its own, so you would just do `getMyList(req);`  Must have fallen in between the cracks? @rast-mk1 
I guess let's leave this as is and let's change this once `getMyList` is fixed.

app/components/Pages/HomePage/index.js
```diff
@@ -18,9 +18,15 @@ export default function HomePage({ listOfCarousels, heroMovie }) {
...
-            </Link>
+            {id === 0 ? (
+              <Link href={`/mylist`} passHref>
+                <CarouselHeader categoryId={id}>{name}</CarouselHeader>
+              </Link>
+            ) : (
+              <Link href={`/category/${id}`} passHref>
+                <CarouselHeader categoryId={id}>{name}</CarouselHeader>
+              </Link>
+            )}
```
wlechowicz, line 29:
```suggestion
            <Link href={id === 0 ? '/mylist' : `/category/${id}`} passHref>
              <CarouselHeader>{name}</CarouselHeader>
            </Link>
```
Additional notes:
1. CarouselHeader does not accept `categoryId` prop
2. CarouselHeader should extend `a`, not `p`. Otherwise, it needs to be wrapper in `<a>` or else the link doesn't work properly.

pages/index.js
```diff
@@ -30,10 +32,13 @@ export async function getServerSideProps({ req }) {
...
     const popularList = await getListOfPopularMovies(req);
     const resultsOfPopularList = popularList.results;
     const popularCategory = { id: 1, name: 'Popular', listOfMovies: popularList.results };
-    const listOfCarousels = [popularCategory, ...listOfCategories];
+    const myListCategory = { id: 0, name: 'My List', listOfMovies: myList };
+    const listOfCarousels = [myListCategory, popularCategory, ...listOfCategories];
```
wlechowicz, line 41: Filter out the whole carousel if My List is empty, for example like this:
```suggestion
    const listOfCarousels = myList?.length
      ? [{ id: 0, name: 'My List', listOfMovies: myList }, popularCategory, ...listOfCategories]
      : [popularCategory, ...listOfCategories];
```

wlechowicz (changes requested):

pages/index.js
rast-mk1: Seems like it did.

wlechowicz (approved):

PR #114, opened 2022-04-06
feat: change username and password backend #1933 

rast-mk1:


hiimsagan (approved):

wlechowicz (changes requested):
Endpoints should never include words like "set", "edit", "get" or "delete", because that's explicitly defined by the HTTP method. PUT itself means "add or modify" therefore it should be just PUT /api/users/

weronika299 (approved):

rast-mk1:
Note to self: `validateSignUpUserName` and `validateSignUpUserPassword` before connecting to DB.

PR #119, opened 2022-04-10
feat: change password view #1932

wolnio:


app/components/Account/Account.styles.js
```diff
@@ -81,3 +81,12 @@ export const Header = styled.h1`
     margin-bottom: none;
   }
 `;
+
+export const ChangePass = styled.span`
```
wlechowicz, line 85: You should use an anchor or a button instead of imitating a link with a span, this is a bad practice.
Since it doesn't navigate away, it should be a button.

app/components/ChangePassword/ChangePassword.styles.js
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components';
+
+const Wrapper = styled.div`
+  width: 100%;
+  padding: 1em 0;
+`;
+
+export { Wrapper };
```
wlechowicz, line 8:
```suggestion
export const Wrapper = styled.div`
  width: 100%;
  padding: 1em 0;
`;
```

app/components/ChangePassword/ChangePassword.stories.js
```diff
@@ -0,0 +1,12 @@
...
+export default {
+  title: 'Account/ChangePassword',
+  component: ChangePassword,
+};
+
+const Template = (args) => {
+  return <ChangePassword {...args} />;
+};
+
+export const Responsive = Template.bind();
```
wlechowicz, line 12:
```suggestion
export const Responsive = Template.bind({});
```

app/components/ChangePassword/ChangePassword.styles.js
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components';
+
+const Wrapper = styled.div`
+  width: 100%;
```
wlechowicz, line 4: `div` is a block element, therefore it's always 100% wide, so this is redundant.

app/components/ChangePassword/index.js
```diff
@@ -0,0 +1,41 @@
+import { Input } from 'components/Form';
+import { Wrapper } from './ChangePassword.styles';
+import { useState } from 'react';
+
+const initialState = Object.freeze({ currentPass: '', newPass: '', repeatPass: '' });
+
+export default function ChangePassword() {
+  const [inputValues, setInputValues] = useState(initialState);
```
wlechowicz, line 8: Add validation and prepare payload for later.

wlechowicz (changes requested):
Buttons need some air between them:
obraz

weronika299 (changes requested):
What do you think about adding an underline or something to ChangePass to more visually emphasize that it is a link?

weronika299:
> > What do you think about adding an underline or something to ChangePass to more visually emphasize that it is a link?
> 
> You mean to make underline visible all the time, not only on hover?

Maybe underlining is not the best, hmm but for example higher font-size and font-weight value? But it's only suggestion so you can try it and decide. ;)

app/components/ChangePassword/ChangePassword.styles.js
wolnio: well, if I remove `width: 100%` the view looks like this: 
image

app/components/ChangePassword/index.js
```diff
@@ -0,0 +1,41 @@
+import { Input } from 'components/Form';
+import { Wrapper } from './ChangePassword.styles';
+import { useState } from 'react';
```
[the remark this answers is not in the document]
wolnio: After review changes I don't use `useState` in that file :)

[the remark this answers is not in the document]
weronika299: Right, but the same situation is in ChangePassword.stories.js file.:)

app/components/Account/Account.styles.js
```diff
@@ -81,3 +82,13 @@ export const Header = styled.h1`
...
+
+export const ChangePass = styled(Button).attrs({ onlyIcon: true })`
+  cursor: pointer;
+  font-weight: bold;
+  &:hover {
+    text-decoration: underline;
+  }
+  margin: 12px;
+  color: #44ec52;
+`;
```
marcelina-hasiak, line 94: Perhaps we could unify these buttons:
Zrzut ekranu 2022-04-14 o 01 28 16

app/components/Account/index.js
```diff
@@ -43,41 +71,63 @@ export default function Account() {
...
-          onlyIcon
-          onClick={toggleIsEditingAvatar}
-          $isEditingAvatar={isEditingAvatar}
-        >
-          <UserAvatar size={200} avatar={selectedAvatar} />
-          <PencilIcon />
-        </EditAvatarButton>
-      </section>
-      {!isEditingAvatar && (
+      {!isChangingPass && (
```
marcelina-hasiak, line 74: I requested some changes in Lena's PR regarding to this logic. Perhaps you could separate it so one block is rendered when isEditingAvatar = true, second when isChangingPass = true, third will be onNameChange = true and finally when nothing is currently edited default jsx appears.  Please check if it's possible to do ;)

marcelina-hasiak (changes requested):

app/components/Account/Account.styles.js
```diff
@@ -31,6 +31,7 @@ export const FlexRow = styled.div`
   flex-direction: row;
   justify-content: space-evenly;
   align-items: center;
+  min-width: 375px;
```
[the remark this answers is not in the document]
wolnio: On a specific device or on all of them?

PR #120, opened 2022-04-12
feat: add MyList button #1901 

marcelina-hasiak:
-Every interested page preloads myList data through serverSideProps. 
-MyList is saved in global state and passed down with Context. 
-LikeButton reads the Context and check if movieId is on the list so it can update its state


rast-mk1:
Nicely done, so why is it still a draft?

marcelina-hasiak:
Thank you @rast-mk1 but I'm still fighting with storybook. I think the problem is MovieCard component which pass its props (id) to LikeButton and then everything explodes ^^'

PR #123, opened 2022-04-13
feat: home page roadblock #1745

weronika299:


app/components/Pages/StartPage/StartPage.styles.js
```diff
@@ -0,0 +1,91 @@
...
+  ${device.tablet} {
+    font-size: 1.8rem;
+  }
+`;
+
+export const SignUpLink = styled.a`
+  padding-left: 10px;
+  font-weight: 700;
+  color: #616161;
+`;
```
lenasokolova, line 91: don't you think SignUpLink should look more like a button? 
image

app/components/Pages/StartPage/index.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default function StartPage() {
+  return (
+    <StartPageContainer>
+      <BackgroundWrapper>
+        <Container>
+          <ContentContainer>
+            <WelcomeHeader>Welcome to inTiVi!</WelcomeHeader>
+            <WelcomeContent>
+              Register now in inTiVi now and enjoy watching the latest movies.
```
rast-mk1, line 20: Word 'now' is repeated.

rast-mk1 (changes requested):
1. There is second scrollbar when window size is reduced. Also, text goes to left side.
obraz
2. There is horizontal scrollbar when browser is set to imitate smartphone.
obraz

app/components/Pages/StartPage/index.js
weronika299: > 1. There is second scrollbar when window size is reduced. Also, text goes to left side.
>    obraz
> 2. There is horizontal scrollbar when browser is set to imitate smartphone.
>    obraz

This horizontal scrollbar appears below 375px width. I removed second vertical scrollbar but text on the left side on tablets is intentional because on tablets  we have more free space on the left side than on the right side when we use this background image.

weronika299:
> 1. There is second scrollbar when window size is reduced. Also, text goes to left side.
>    obraz
> 2. There is horizontal scrollbar when browser is set to imitate smartphone.
>    obraz

This horizontal scrollbar appears below 375px width. I removed second vertical scrollbar but text on the left side on tablets is intentional because on tablets we have more free space on the left side than on the right side when we use this background image.

app/components/Pages/StartPage/StartPage.styles.js
```diff
@@ -0,0 +1,83 @@
...
+export const ContentContainer = styled.div`
+  min-width: 200px;
+  max-width: 100%;
+  text-align: center;
+  ${device.tablet} {
+    width: 400px;
+  }
+`;
+
+export const WelcomeHeader = styled.p`
```
marcelina-hasiak, line 57: If it's header wrap the content with catchy H1 tag ;)

marcelina-hasiak (changes requested):
Small adjustment needed, apart from that good job ;)

app/components/Pages/StartPage/StartPage.styles.js
weronika299: Okay, I will change it.:)

app/components/Pages/StartPage/StartPage.styles.js
weronika299: Okay, I will change it. Thanks!

rast-mk1 (approved):

lenasokolova (approved):
in my opinion, now it's a lot better :) nicely done!

rast-mk1 (approved):

weronika299:
> in my opinion, now it's a lot better :) nicely done!



> in my opinion, now it's a lot better :) nicely done!

Yes, you're right ;)

weronika299:
> in my opinion, now it's a lot better :) nicely done!



> in my opinion, now it's a lot better :) nicely done!

Yes, you're right ;)

PR #124, opened 2022-04-13
feat: change username view #1931

lenasokolova:


rast-mk1:
What do you think about having `cursor: pointer` when hovering over username?

app/components/Account/index.js
```diff
@@ -43,42 +56,88 @@ export default function Account() {
...
+            />
+            <FlexRow>
+              <Button primary onClick={() => alert('Username will be updated soon')}>
+                Save
+              </Button>
+              <Button onClick={() => setIsChangingUsername(false)}>Cancel</Button>
+
+              {/* add onClick to both Buttons */}
+            </FlexRow>
+          </section>
```
marcelina-hasiak, line 134: Hmmm... If it's possible I'd omit logic nesting it's quite unreadable, what's more I see duplicated section with EditUsername perhaps you'll fing a way to improve it ;)

app/components/Account/index.js
```diff
@@ -43,42 +56,88 @@ export default function Account() {
...
+            >
+              <UserAvatar size={200} avatar={selectedAvatar} />
+              <PencilIcon />
+            </EditAvatarButton>
+          </section>
+          {!isEditingAvatar && (
+            <>
+              <section>
+                <EditUsername onClick={() => setIsChangingUsername(!isChangingUsername)}>
+                  {user.name && <Username>{user.name}</Username>}
```
marcelina-hasiak, line 75: Pass callback here just like you did inside PencilIcon. But wait, do you need both event handlers?

marcelina-hasiak (changes requested):

rast-mk1 (approved):

